PR #15, opened 2020-06-16
support adding packages using git url

ahelsayd:


jumpscale/servers/threebot/threebot.py
```diff
@@ -231,11 +238,20 @@ def get_packages(self):
     def list_all(self):
         return self.packages.keys()
 
-    def add(self, path):
+    def add(self, path: str = None, giturl: str = None):
         # TODO: Check if package already exists
+
+        if not (path or giturl):
```
xmonader, line 244: what happens in case of both were provided?

is suggest `any(...) and not all(...)`

Issue #16, opened 2020-06-16
Error in start Nginx when `443` port is used in package locations


dmahmouali:
![image](https://user-images.githubusercontent.com/36021484/84789377-74835f80-aff0-11ea-9088-66780f4babdc.png)

xmonader:
@alichaddad does that mean we don't use a self signed in case of private ip?

Dina-Abd-Elrahman:
- verified.

PR #22, opened 2020-06-17
fail back to self-signed certs if certbot failed

m-motawea:
issue: https://github.com/threefoldtech/js-sdk/issues/16

### DIFF
--- a/jumpscale/sals/nginx/nginx.py
+++ b/jumpscale/sals/nginx/nginx.py
@@ -119,14 +119,11 @@ class Website(Base):
         )
         if res[0] != 0:
             if self.selfsigned:
-                res = self.generate_self_signed_certificates()
-                if res and res[0] != 0:
-                    raise j.exceptions.JSException(f"Failed to generate self-signed certificate.{res}")
+                self.generate_self_signed_certificates()
                 j.sals.fs.write_file(self.cfg_file, self.get_config())
                 return res
             else:
                 raise j.exceptions.JSException(f"Failed to generate certificates by certbot.{res}")
-        return res
 
     def generate_self_signed_certificates(self):
         self.selfsigned = True
@@ -135,7 +132,8 @@ class Website(Base):
         res = j.sals.process.execute(
             f"openssl req -nodes -x509 -newkey rsa:4096 -keyout {self.parent.cfg_dir}/key.pem -out {self.parent.cfg_dir}/cert.pem -days 365 -subj '/CN=localhost'"
         )
-        return res
+        if res[0] != 0:
+            raise j.exceptions.JSException(f"Failed to generate self-signed certificate.{res}")
 
     def configure(self, generate_certificates=True):
         j.sals.fs.mkdir(self.cfg_dir)
--- a/jumpscale/sals/nginx/nginx.py
+++ b/jumpscale/sals/nginx/nginx.py
@@ -86,7 +86,7 @@ class Website(Base):
     port = fields.Integer(default=80)
     locations = fields.Factory(Location)
     letsencryptemail = fields.String()
-    selfsigned = fields.Boolean(default=False)
+    selfsigned = fields.Boolean(default=True)
 
     @property
     def cfg_dir(self):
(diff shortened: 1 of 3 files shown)


jumpscale/sals/nginx/nginx.py
```diff
@@ -114,26 +115,48 @@ def get_config(self):
 
     def generate_certificates(self):
         if self.ssl:
-            j.sals.process.execute(
+            res = j.sals.process.execute(
```
xmonader, line 118: how would you know if it was a success? it returns a tuple of (rc, out, err). you need to explicitly check there and raise if needed no? and die is set to False by default already so you wouldn't know.

jumpscale/sals/nginx/nginx.py
```diff
@@ -114,26 +115,48 @@ def get_config(self):
...
+            return res
+
+    def generate_self_signed_certificates(self):
+        if self.ssl:
+            self.selfsigned = True
+            if j.sals.fs.exists(f"{self.parent.cfg_dir}/key.pem") and j.sals.fs.exists(
+                f"{self.parent.cfg_dir}/cert.pem"
+            ):
+                return
+            res = j.sals.process.execute(
```
xmonader, line 130: same comment above

jumpscale/sals/nginx/nginx.py
```diff
@@ -114,26 +115,48 @@ def get_config(self):
...
+        self.selfsigned = False
         j.sals.fs.write_file(self.cfg_file, self.get_config())
-
-        if generate_certificates:
-            self.generate_certificates()
+        self.selfsigned = failback
+
+        if generate_certificates and self.ssl:
+            res = self.generate_certificates()
+            if res and res[0] != 0:
```
xmonader, line 147: we should be clear about what is res[0] in that case I had to read the code above to figure out what is intended

jumpscale/sals/nginx/nginx.py
```diff
@@ -114,26 +115,48 @@ def get_config(self):
...
+                if failback:
+                    # failed to generate cert by certbot
+                    # use self_signed certs
+                    self.generate_self_signed_certificates()
```
xmonader, line 151: Here we assume this call always works without problems and ignoring the rc, out, err data

jumpscale/sals/nginx/nginx.py
```diff
@@ -113,27 +114,46 @@ def get_config(self):
...
+        return res
+
+    def generate_self_signed_certificates(self):
+        self.selfsigned = True
+        if j.sals.fs.exists(f"{self.parent.cfg_dir}/key.pem") and j.sals.fs.exists(f"{self.parent.cfg_dir}/cert.pem"):
+            return
+        res = j.sals.process.execute(
+            f"openssl req -nodes -x509 -newkey rsa:4096 -keyout {self.parent.cfg_dir}/key.pem -out {self.parent.cfg_dir}/cert.pem -days 365 -subj '/CN=localhost'"
+        )
+        return res
```
grimpy, line 138: wouldnt it make more sense this throws it's own excpetion on failure?

jumpscale/sals/nginx/nginx.py
```diff
@@ -113,27 +114,46 @@ def get_config(self):
...
+        if res[0] != 0:
+            if self.selfsigned:
+                res = self.generate_self_signed_certificates()
+                if res and res[0] != 0:
+                    raise j.exceptions.JSException(f"Failed to generate self-signed certificate.{res}")
+                j.sals.fs.write_file(self.cfg_file, self.get_config())
+                return res
+            else:
+                raise j.exceptions.JSException(f"Failed to generate certificates by certbot.{res}")
+        return res
```
grimpy, line 129: return is not really needed as you throw exception on failure
you might want to consider returning a boolean indicating if it's selfsigned or generated by certbot (document in docstring if you decide to do so)

jumpscale/sals/nginx/nginx.py
```diff
@@ -113,27 +114,46 @@ def get_config(self):
...
     def configure(self, generate_certificates=True):
         j.sals.fs.mkdir(self.cfg_dir)
-
         for location in self.get_locations():
             location.configure()
 
+        failback = self.selfsigned
+        self.selfsigned = False
         j.sals.fs.write_file(self.cfg_file, self.get_config())
+        self.selfsigned = failback
```
grimpy, line 148: i think you need to write the file after generating certs so you know if it's selfsigned or generated by certbot

Issue #35, opened 2020-06-18
add package by git url faces rate limits


xmonader:
let's clone the whole repo of the package instead

Dina-Abd-Elrahman:
- verified

PR #39, opened 2020-06-18
clone repo when adding package using the giturl instead of using the api

ahelsayd:
see https://github.com/threefoldtech/js-sdk/issues/35

### DIFF
--- a/jumpscale/servers/threebot/threebot.py
+++ b/jumpscale/servers/threebot/threebot.py
@@ -213,19 +213,12 @@ class PackageManager(Base):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._threebot = None
-        self._github = None
 
     @property
     def threebot(self):
         if self._threebot is None:
             self._threebot = j.servers.threebot.get(self.instance_name)
         return self._threebot
-
-    @property
-    def github(self):
-        if self._github is None:
-            self._github = j.clients.github.get("default")
-        return self._github
 
     def get(self, package_name):
         package_path = self.packages.get(package_name)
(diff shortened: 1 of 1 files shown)


jumpscale/servers/threebot/threebot.py
```diff
@@ -241,13 +234,21 @@ def list_all(self):
     def add(self, path: str = None, giturl: str = None):
         # TODO: Check if package already exists
 
-        if any([path, giturl]) and not all([path, giturl]):
+        if not any([path, giturl]) or all([path, giturl]):
```
xmonader, line 237: should be `if not any([path, giturl]) and not all([path, giturl])`

we want path or giturl to be there but not both of them

jumpscale/servers/threebot/threebot.py
```diff
@@ -241,13 +234,21 @@ def list_all(self):
...
             raise j.exceptions.Value("either path or giturl is required")
 
         if giturl:
-            org, repo, _, branch, path = urlparse(giturl).path.lstrip("/").split("/", 4)
-            repo = self.github.get_repo(f"{org}/{repo}")
-            path = repo.download_directory(path, DOWNLOADED_PACKAGES_PATH, branch=branch)
+            url = urlparse(giturl)
+            org, repo, _, branch, package_path = url.path.lstrip("/").split("/", 4)
```
xmonader, line 242: can we validate the number of components before proceeding?

jumpscale/servers/threebot/threebot.py
waleedhammam: we need the condition to be the opposite, if it enteres the if it'll raise not continue

thanks @abom

ahelsayd: i think it is correct
raise error if  not any of path / giturl passed or if both of them are passed

Issue #31, opened 2020-06-18
protect /admin url


xmonader:
make sure /admin is protected unless in development mode.

Dina-Abd-Elrahman:
- verified.

PR #43, opened 2020-06-18
Development locations auth

m-motawea:
issue: https://github.com/threefoldtech/js-sdk/issues/31


jumpscale/sals/nginx/templates/website.conf
```diff
@@ -14,6 +14,10 @@
     {% if website.domain %}
         server_name                 ~^(www\.)?{{ website.domain }}$;
     {% endif %}
+        error_page 401 = @error401;
```
abom, line 17: I think we need to deal with 403 too, see admin_only, it can return 403:

https://github.com/threefoldtech/js-sdk/blob/development_locations_auth/jumpscale/packages/auth/bottle/auth.py#L222

Issue #62, opened 2020-06-21
Code server installation on host


AhmedSa-mir:
Code server should be installed on adding the package on host

Dina-Abd-Elrahman:
- verified.

PR #63, opened 2020-06-21
Added codeserver installation on adding package

AhmedSa-mir:
See https://github.com/threefoldtech/js-sdk/issues/62

### DIFF
--- a/jumpscale/packages/codeserver/package.py
+++ b/jumpscale/packages/codeserver/package.py
@@ -1,9 +1,11 @@
 from jumpscale.god import j
+import os
 
 
 class codeserver:
     def __init__(self):
         self.bin_path = j.sals.fs.join_paths(j.core.dirs.BINDIR, "code-server")
+        self.script_path = j.sals.fs.join_paths(os.path.dirname(__file__), "codeserver-install.sh")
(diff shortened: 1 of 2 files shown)


jumpscale/packages/codeserver/package.py
```diff
@@ -1,9 +1,11 @@
 from jumpscale.god import j
+import os
 
 
 class codeserver:
     def __init__(self):
         self.bin_path = j.sals.fs.join_paths(j.core.dirs.BINDIR, "code-server")
+        self.script_path = j.sals.fs.join_paths(os.path.dirname(__file__), "codeserver-install.sh")
```
waleedhammam, line 8: I think calling it in this `__init__` function is not OK ? It may cause the package to corrupt and cause deletion error also.
 maybe in install method or `__init__.py` if you wanna call it in docker file and here too, will be something like  `j.packages.codeserver.install()`

AhmedSa-mir: I don't call anything here. I just get the path for the script!

waleedhammam: okay then fine (y)

Issue #12, opened 2020-06-15
Packages URLs issues


dmahmouali:
- [x] have to use `/` after `/admin` to work

- [x] have to remove `/` from the end of `tfgrid_solutions/chats/<chat>/` to work

abom:
https://github.com/threefoldtech/js-sdk/pull/6/commits/2ba9eedcb6097d7a2894ef976bfef6cfe96b3183

dmahmouali:
verifying

PR #75, opened 2020-06-22
fix nginx locations trailing slash

dmahmouali:
issue: https://github.com/threefoldtech/js-sdk/issues/12

### DIFF
--- a/jumpscale/sals/nginx/templates/location.conf
+++ b/jumpscale/sals/nginx/templates/location.conf
@@ -1,4 +1,4 @@
-location {{ location.path_url }} {
+location {{ location.path_url }} {{ "/" if not location.path_url.endswith("/") }} {
 
     # to keep hostname (if it's with e.g. port)
     absolute_redirect off;
--- a/jumpscale/sals/nginx/templates/location.conf
+++ b/jumpscale/sals/nginx/templates/location.conf
@@ -1,4 +1,4 @@
-location {{ location.path_url }} {{ "/" if not location.path_url.endswith("/") }} {
+location {{ location.path_url }} {
 
     # to keep hostname (if it's with e.g. port)
     absolute_redirect off;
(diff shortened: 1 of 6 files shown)


jumpscale/servers/threebot/threebot.py
```diff
@@ -54,7 +54,7 @@ def default_config(self):
                     "name": bottle_server.get("name"),
                     "host": bottle_server.get("host"),
                     "port": bottle_server.get("port"),
-                    "path_url": j.sals.fs.join_paths(self.package.base_url, bottle_server.get("path_url").lstrip("/")),
```
ahelsayd, line 57: we need the lstrip 
all paths passed to path.join (except the first one) shouldn't starts with `/`

jumpscale/sals/nginx/templates/location_static.conf
```diff
@@ -1,6 +1,6 @@
-    alias  {{ location.path_location }}{{ "/" if not location.path_location.endswith("/") }};
```
ahelsayd, line 1: make sure in threebot that we do rstrip("/") for `path_location` and add / here at the end

jumpscale/sals/nginx/templates/location_proxy.conf
```diff
@@ -11,4 +11,4 @@
     proxy_set_header X-Real-IP $remote_addr;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
-    proxy_pass {{ location.scheme }}://{{ location.host }}:{{ location.port }}{{ location.path_dest }}{{ "/" if not location.path_dest.endswith("/") }};
```
ahelsayd, line 14: make sure in threebot that we do `rstrip("/")` for `path_dest` and add `/` here at the end

jumpscale/sals/nginx/templates/location.conf
```diff
@@ -1,4 +1,4 @@
-location {{ location.path_url }} {{ "/" if not location.path_url.endswith("/") }} {
```
ahelsayd, line 1: make sure in threebot that we do `rstrip("/")` for `path_url` and add `/` here at the end

PR #78, opened 2020-06-23
add tool to install, start and stop redis server

ahelsayd:


jumpscale/tools/redis/redis.py
```diff
@@ -0,0 +1,41 @@
...
+
+class RedisServer(Base):
+    name = fields.String(default="redis")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.cmd = j.tools.startupcmd.get(self.name)
+
+    @property
+    def installed(self):
```
xmonader, line 13: this will only work on ubuntu no? i think the sane check would be for redis-server binary in path.
I think we probably should add it for installation instruction on the host, or that's used in case of container only?

Issue #85, opened 2020-06-23
add entry point around starting 3bot


xmonader:
should be as easy as `3bot start`

waleedhammam:
Progress

- Added https://github.com/threefoldtech/js-sdk/commits/development_entry v1 has `threebot start` and `threebot stop`
- Basic identity checks

TODO:

- Add more identity checks (port from jsx sdk)

waleedhammam:
Add more identity checks -> added

hossnys:
verified

PR #90, opened 2020-06-23
Development entry point

waleedhammam:
for issue: https://github.com/threefoldtech/js-sdk/issues/85

- Add threebot start and threebot stop


waleedhammam:
> start with a sideeffect of registering an identity? This does not feel right to me, should just take an optional identity name if none is passed the default should be used
> 
> If none is configured exit none 1 with errror message

this was unfinished yet ;)

jumpscale/entry_points/threebot.py
```diff
@@ -0,0 +1,71 @@
...
+    if identity:
+        if j.core.identity.find(identity):
+            print(f"Setting default identity to {identity}\nStarting threebot server...")
+            j.core.identity.set_default(identity)
+            j.core.identity.me.save()
+        else:
+            j.tools.console.printcolors(
+                "{RED}Identity %s is not set, please configure it or start with the default one" % identity
+            )
+            return
```
grimpy, line 32: sys.exit(1) ?

jumpscale/entry_points/threebot.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if not j.core.identity.list_all():
+        identity_data = IdentityManager()
+        identity, email, words, explorer = identity_data.ask_identity()
```
grimpy, line 36: this returns a named tuple now we can use it as such instead of unpacking here

jumpscale/entry_points/threebot.py
```diff
@@ -0,0 +1,71 @@
...
+        me.register()
+        me.save()
+
+    cmd = j.tools.startupcmd.get("threebot_default")
+    cmd.start_cmd = "jsng 'j.servers.threebot.start_default(wait=True)'"
+    cmd.process_strings_regex = [j.tools.nginx.get("default").check_command_string, "nginx.*"]
+    cmd.ports = [8000, 8999]
+    cmd.start()
+    print("\n✅ Threebot server started\n")
+    j.tools.console.printcolors("{WHITE}Visit admin dashboard at: {GREEN}http://localhost/\n")
```
grimpy, line 49: if we print this like that how would this look if invoked from 3sdk, where they actually have to access the admin portal on another port :-S? @xmonader suggestions?

jumpscale/entry_points/threebot.py
```diff
@@ -0,0 +1,71 @@
+import click
+
+from jumpscale.loader import j
+from jumpscale.threesdk.identitymanager import IdentityManager
+
+NETWORKS = {"mainnet": "explorer.grid.tf", "testnet": "explorer.testnet.grid.tf", "devnet": "explorer.devnet.grid.tf"}
```
grimpy, line 6: this variable is not used

jumpscale/entry_points/threebot.py
xmonader: problem is it was consistent in jsx because it was always 4000 and used via port forward so maybe we can do port check? as a simple check or check for startup_cmd running on the host namespace? if that the case we can show the localhost message?

waleedhammam: will only print in host

Issue #48, opened 2020-06-18
If the last network solution with the same name is canceled the other ones are irrelevant


dmahmouali:
in call `get_solutions(Network)` if the last network solution with the same name is **canceled** the other ones are irrelevant, It should be ignored so.

ranatrk:
**Approach tried:(will be disgarded)** Get all networks with this name from deleted and deployed networks and check if it was deleted then this network is not considered, if it is still deployed then it is considered. 
this approach is too slow as by time the number of total reservations will increase and considering deleted reservations is not effiecient

**Suggested approach**: After discussing with @grimpy  and @zaibon, the suggested approach to be implemented is saving the parent of the network being created in the reservation's metadata and so when checking for a network will only look at the parent reservation of it (linked list approach) and so on to know if the network is deployed or deleted

BolaNasr:
**Verified**
create network -> create 2 ubuntu conatiner 
Delete network -> refresh multiple time and close and open server ( still deleted )

PR #91, opened 2020-06-23
Development network solutions

ranatrk:
- Add parent_network to metadata of any reservation that adds nodes to an existing network
- when deleting a network from admin panel that has a parent network(or more), all parent networks are cancelled from the explorer as well
- fix use of DistType .SSD enum value and name in ubuntu solution deploy

https://github.com/threefoldtech/js-sdk/issues/48

### DIFF
--- a/jumpscale/sals/reservation_chatflow/reservation_chatflow.py
+++ b/jumpscale/sals/reservation_chatflow/reservation_chatflow.py
@@ -951,22 +951,26 @@ Deployment will be cancelled if it is not successful {remaning_time}
             if network.name == name:
                 return Network(network, expiration, bot, reservations, currency, resv_id)
 
-    def list_networks(self, tid, reservations=None):
+    def list_networks(self, tid, reservations=None, include_deleted=False):
         """list all available networks from reservations
 
         Args:
             tid (int): user tid
             reservation (list of jumpscale.clients.explorer.models.TfgridWorkloadsReservation1): list of reservation objects
+            include_deleted (boolean): include deleted networks
 
         Returns:
             [type]: [description]
         """
         if not reservations:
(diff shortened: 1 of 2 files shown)


jumpscale/sals/reservation_chatflow/reservation_chatflow.py
```diff
@@ -957,12 +976,13 @@ def list_networks(self, tid, reservations=None):
         Args:
             tid (int): user tid
             reservation (list of jumpscale.clients.explorer.models.TfgridWorkloadsReservation1): list of reservation objects
+            include_deleted (boolean): include deleted networks
```
grimpy, line 979: remove docstring of unexisting arg

jumpscale/sals/reservation_chatflow/reservation_chatflow.py
```diff
@@ -116,7 +129,13 @@ def update(self, tid, currency=None, bot=None):
                 j.sals.reservation_chatflow.wait_payment(
                     bot, rid, threebot_app=True, reservation_create_resp=reservation_create
                 )
-            return self._sal.wait_reservation(self._bot, rid)
+            wait_reservation_results = self._sal.wait_reservation(self._bot, rid)
+            # Update solution saved locally
+            explorer_name = self._sal._explorer.url.split(".")[1]
+            solution = self.solutions.get(f"{explorer_name}_{self.resv_id}")
```
grimpy, line 135: it feels like this would work only during the first update
as next time the self.rev_id will be the updated reservation id.

I would suggest deleting the old network solution and creating a new one with the correct name

when this is done you can also update `self.resv_id = rid`

ranatrk: Yes it would only work once, apparently when I tested it, the updates were done already from the explorer so it wasn't reflected on multiple child networks.

PR #100, opened 2020-06-24
sanitization check if threebot server is alredy started

waleedhammam:


jumpscale/entry_points/threebot.py
```diff
@@ -44,17 +46,21 @@ def start(identity=None):
...
         me.save()
 
-    ports = [8999, 80, 443, 16000]
     used_ports = []
-    for port in ports:
-        if j.sals.nettools.tcp_connection_test(ipaddr="localhost", port=port, timeout=1):
-            used_ports.append(port)
+    ports_error_msg = ""
+    for service_name, service_port in SERVICES_PORTS.items():
+        if j.sals.nettools.tcp_connection_test(ipaddr="localhost", port=service_port, timeout=1):
```
grimpy, line 52: it would be better if you use psutils.connect() ans check if you have those ports in listening state, because they might be bind to other ips than localhost so localhost test will fail but the port isnt free

Issue #103, opened 2020-06-25
when starting threebot: cannot find static dir /home/ahmed/sandbox/code/github/jumpscaleX_weblibs/static


xmonader:
```

Traceback (most recent call last):
> File "/home/ahmed/.cache/pypoetry/virtualenvs/js-sdk--OGPzDz5-py3.6/bin/jsng", line 33, in <module>
    sys.exit(load_entry_point('js-ng', 'console_scripts', 'jsng')())
    │   │    └ <function importlib_load_entry_point at 0x7f2209b78e18>
    │   └ <built-in function exit>
    └ <module 'sys' (built-in)>

  File "/home/ahmed/.cache/pypoetry/virtualenvs/js-sdk--OGPzDz5-py3.6/src/js-ng/jumpscale/entry_points/jsng.py", line 24, in run
    sys.exit(print(eval(sys.argv[1])))
    │   │               │   └ ['/home/ahmed/.cache/pypoetry/virtualenvs/js-sdk--OGPzDz5-py3.6/bin/jsng', 'j.servers.threebot.start_default(wait=True)']
    │   │               └ <module 'sys' (built-in)>
    │   └ <built-in function exit>
    └ <module 'sys' (built-in)>

  File "<string>", line 1, in <module>

  File "/home/ahmed/wspace/js/js-sdk/jumpscale/servers/threebot/__init__.py", line 14, in start_default
    server.start(wait=wait)
    │      │     │    └ True
    │      │     └ True
    │      └ <function ThreebotServer.start at 0x7f220116b0d0>
    └ ThreebotServer(_Base__instance_name='default', _Base__parent=None, ___package_manager=<jumpscale.core.base.factory.jumpscales...

  File "/home/ahmed/wspace/js/js-sdk/jumpscale/servers/threebot/threebot.py", line 524, in start
    self.packages.install(package)
    │    │                └ <jumpscale.servers.threebot.threebot.Package object at 0x7f21faea3198>
    │    └ <property object at 0x7f2201167548>
    └ ThreebotServer(_Base__instance_name='default', _Base__parent=None, ___package_manager=<jumpscale.core.base.factory.jumpscales...

  File "/home/ahmed/wspace/js/js-sdk/jumpscale/servers/threebot/threebot.py", line 373, in install
    raise j.exceptions.NotFound(f"Cannot find static dir {path}")
          │ └ <property object at 0x7f2203fa3318>
[...]

PR #109, opened 2020-06-25
handled cloning weblibs failure

dmahmouali:
issue: https://github.com/threefoldtech/js-sdk/issues/103

### DIFF
--- a/jumpscale/packages/weblibs/package.py
+++ b/jumpscale/packages/weblibs/package.py
@@ -16,9 +16,12 @@ class weblibs:
                 try:
                     j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch, depth=1)
                     return
-                except Exception:
+                except Exception as e:
                     retries -= 1
-            raise j.exceptions.Timeout("Clone weblibs repo failed")
+                    msg = str(e)
+                    # check if error not lost internet connection don't try again
+                    if not retries or msg.find("Could not resolve host") == -1:
+                        raise e
 
     def uninstall(self):
         """Called when package is deleted
--- a/jumpscale/packages/weblibs/package.py
+++ b/jumpscale/packages/weblibs/package.py
@@ -11,15 +11,14 @@ class weblibs:
         """Called when package is added
         """
         if not j.sals.fs.exists(j.sals.fs.join_paths(self.path, "jumpscaleX_weblibs")):
-            retries = 2
+            retries = 3
             while retries:
                 try:
                     j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch, depth=1)
-                    break
+                    return
                 except Exception:
-                    if retries == 1:
-                        raise j.exceptions.Timeout("Clone weblibs repo failed")
                     retries = retries - 1
+            raise j.exceptions.Timeout("Clone weblibs repo failed")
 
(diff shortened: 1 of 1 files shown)


jumpscale/packages/weblibs/package.py
```diff
@@ -11,7 +11,15 @@ def install(self):
         """Called when package is added
         """
         if not j.sals.fs.exists(j.sals.fs.join_paths(self.path, "jumpscaleX_weblibs")):
-            j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch)
+            retries = 5
```
waleedhammam, line 14: 5 is too much let's make it 2 :D

jumpscale/packages/weblibs/package.py
```diff
@@ -11,7 +11,15 @@ def install(self):
...
+            while retries:
+                try:
+                    j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch)
```
waleedhammam, line 17: can make a shallow cloning to reduce cloning size depth=1 for example

jumpscale/packages/weblibs/package.py
waleedhammam, line 14: 5 is too much let's make it 3

jumpscale/packages/weblibs/package.py
dmahmouali: added, thanks

jumpscale/packages/weblibs/package.py
xmonader: what happens if it failed and the dir got created and not a real git repo?

waleedhammam: git doesn't create the repo file unless it's completed. no ?

jumpscale/packages/weblibs/package.py
```diff
@@ -11,7 +11,15 @@ def install(self):
...
         """
         if not j.sals.fs.exists(j.sals.fs.join_paths(self.path, "jumpscaleX_weblibs")):
-            j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch)
+            retries = 2
+            while retries:
+                try:
+                    j.tools.git.clone_repo(url=self.url, dest=self.path, branch_or_tag=self.branch, depth=1)
+                    break
+                except Exception:
+                    if retries == 1:
```
waleedhammam, line 20: if retries:

dmahmouali: but what for while condition ?

jumpscale/packages/weblibs/package.py
```diff
@@ -11,7 +11,15 @@ def install(self):
...
+                        raise j.exceptions.Timeout("Clone weblibs repo failed")
+                    retries = retries - 1
```
waleedhammam, line 22: retries -= 1

jumpscale/packages/weblibs/package.py
waleedhammam: let's make it after while then ?

jumpscale/packages/weblibs/package.py
dmahmouali: tried cloning with wrong branch and anther with no internet connection and no dirs created :+1:

Issue #94, opened 2020-06-24
Missing options in wallet manager in admin panel


ranatrk:
- When creating a wallet from the admin panel, the network type is always TEST, it should either be chosen by the user or set according to the current identity's explorer instance (**Update**: Add option to create new identity/switch between identities)
- Can't add trust lines to existing wallets from the admin panel

ranatrk:
The problem is when switching the explorer to main from the admin, the identity used is not switched to the identity used in explorer, and the default explorer is still testnet.
Looking into that now

ranatrk:
Based on the comment in the PR https://github.com/threefoldtech/js-sdk/pull/118#issuecomment-650101467, will add an option to setup identities from the admin dashboard to be able to switch between identities instead of explorers in the dashboard

ranatrk:
Based on the comment in the PR https://github.com/threefoldtech/js-sdk/pull/118#issuecomment-650101467, will add an option to setup identities from the admin dashboard to be able to switch between identities instead of explorers in the dashboard

ranatrk:
updates : https://github.com/threefoldtech/js-sdk/pull/261

ranatrk:
Identities
- Add/register new identity from dashboard settings
- View identities in dashboard settings
- Delete identity from dashboard
- Change default identity used from dashboard (will be set in ` j.core.identity.me`)

Wallets
- Add update known trustline option for wallets from dashboard

hossnys:
verified :

add new identity : 

![Screenshot from 2020-07-26 12-50-43](https://user-images.githubusercontent.com/26298009/88477697-70792480-cf42-11ea-9ae2-b60214e94840.png)

view identities :
![Screenshot from 2020-07-26 13-23-17](https://user-images.githubusercontent.com/26298009/88477735-b46c2980-cf42-11ea-8449-a2ba70e07e5b.png)


delete identity :

![Screenshot from 2020-07-26 12-51-20](https://user-images.githubusercontent.com/26298009/88477699-74a54200-cf42-11ea-8b1e-1fb4a0b04a68.png)


- also changing identity from dashboard is working fine.

update trustline option :
![Screenshot from 2020-07-26 13-26-03](https://user-images.githubusercontent.com/26298009/88477775-1462d000-cf43-11ea-8ce4-47fab0fb65cd.png)

PR #118, opened 2020-06-25
Fix switching explorer from admin panel

ranatrk:
- when switching explorer from admin panel, switch the default explorer and the default identity
- change explorer instance in zos sal if the default explorer url is changed

https://github.com/threefoldtech/js-sdk/issues/94

### DIFF
--- a/jumpscale/core/identity/__init__.py
+++ b/jumpscale/core/identity/__init__.py
@@ -126,7 +126,8 @@ class Identity(Base):
                 )
             tid = user.id
         self._tid = tid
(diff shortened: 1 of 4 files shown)


grimpy:
in jsng an identity has concept of explorer... I think it makes more sense to keep it this way and allow the portal to switch between identities (which are linked to an explorer) not switching the explorer itself within an identity cause this same identity might not exist  on another explorer...

In all the code above I see we are trying to get the 'default' explroer instead of the explorer of the current identity j.me.explorer...

imho it should be the other way around...

Issue #117, opened 2020-06-25
threebot entry point improvments


waleedhammam:
- add threebot status [Running - Stopped]

- add threebot restart

hossnys:
verified 
```
root@js-ng:/sandbox/code/github/threefoldtech/js-sdk# threebot status
Server is running

root@js-ng:/sandbox/code/github/threefoldtech/js-sdk# threebot restart
Threebot server Stopped

✅ Threebot server started


```

PR #123, opened 2020-06-28
threebot status/restart

m-motawea:
issue: https://github.com/threefoldtech/js-sdk/issues/117


abom:
Now threebot factory is only doing tea.

PR #130, opened 2020-06-29
fix bugs in Digital Ocean client and update its docs

RafyAmgadBenjamin:
Related to this issue https://github.com/threefoldtech/js-ng/issues/156
Fix bugs in the digital ocean client and update the docs


abom:
Why do we have a field name that ends with underscore? e.g. `token_`

Secret fields are not defined like this, there's a fields type called `Secret`, and the name of the field does not matter.

PR #133, opened 2020-06-29
Add gitea_deploy

BolaNasr:


examplescripts/gitea_deploy.py
```diff
@@ -0,0 +1,308 @@
+import math
+import random, time
+import datetime
+import netaddr
+from jumpscale.god import j
```
grimpy, line 5: outdated should be .loader

Issue #144, opened 2020-06-30
Show all existing packages in admin frontend


AhmedSa-mir:
Currently we show only the installed ones

Dina-Abd-Elrahman:
- verified

PR #148, opened 2020-06-30
Changes in admin package manager

AhmedSa-mir:
- List all packages under j.packages namespace
- Stripped package path
- Some naming refactoring in js
See https://github.com/threefoldtech/js-sdk/issues/144


jumpscale/packages/admin/actors/packages.py
```diff
@@ -1,5 +1,5 @@
 from jumpscale.servers.gedis.baseactor import BaseActor, actor_method
-from jumpscale.loader import j
+from jumpscale.god import j
```
xmonader, line 2: why  switching the loader?

jumpscale/servers/threebot/threebot.py
```diff
@@ -282,7 +282,11 @@ def get(self, package_name):
             return Package(path=package_path, default_domain=self.threebot.domain, default_email=self.threebot.email)
 
     def get_packages(self):
```
xmonader, line 284: that won't work in case if you registered outside package right?

jumpscale/packages/admin/frontend/frontend_src/sources/services/packages.js
```diff
@@ -17,8 +17,8 @@ class PackagesService extends Service {
         return this.getCall("list_packages");
     }
 
-    packagesNames() {
-        return this.getCall("packages_names");
+    getInstalledPackages() {
```
xmonader, line 20: It should be the available packages no?

jumpscale/packages/admin/actors/packages.py
AhmedSa-mir: Didn't change that don't know what happened by merge or something. anyway reverted

jumpscale/servers/threebot/threebot.py
AhmedSa-mir: As per @abom all paths for external packages will be listed under j.packages.__path__

jumpscale/packages/admin/frontend/frontend_src/sources/services/packages.js
AhmedSa-mir: No these are the installed ones only. The available packages are fetched by list_packages()

jumpscale/servers/threebot/threebot.py
```diff
@@ -411,6 +432,24 @@ def install(self, package):
...
+    def reload(self, package_name):
+        package = self.get(package_name)
+        if not package:
+            raise j.exceptions.NotFound(f"{package_name} package not found")
+
+        # re-install package
+        package.install()
+
+        # install package if threebot is started
+        if self.threebot.started:
```
ahelsayd, line 444: move this check up and if threebot is not started raise an error
reload mean the server is started and you need to reload the package without restarting the server
so reloading the package while the server is not running doesn't make sense

PR #154, opened 2020-07-01
Attempt to load chatflow inside the correct package

grimpy:
This allows relative imports inside the chatflow and the package it
belongs into

Signed-off-by: Jo De Boeck <deboeck.jo@gmail.com>


jumpscale/packages/chatflows/actors/chatbot.py
```diff
@@ -60,10 +61,31 @@ def _scan_chats(self, path):
                 name = j.sals.fs.basename(path)[:-3]
```
xmonader, line 61: can use j.sals.fs.stem instead ^_^

Issue #126, opened 2020-06-28
error in j.servers.threebot.default.packages.install_all()


waleedhammam:
```
S-NG> j.servers.threebot.default.packages.install_all()                                              
2020-06-28 20:27:19.247 | ERROR    | jumpscale.tools.errorhandler.errorhandler:_handle_exception:55 - AttributeError: 'NoneType' object has no attribute 'load'
Traceback (most recent call last):
  File "/home/waleed/.cache/pypoetry/virtualenvs/js-sdk-6NoanB64-py3.8/bin/jsng", line 33, in <module>
    sys.exit(load_entry_point('js-ng', 'console_scripts', 'jsng')())
    │   │    └ <function importlib_load_entry_point at 0x7fc5f677c1f0>
    │   └ <built-in function exit>
    └ <module 'sys' (built-in)>

  File "/home/waleed/.cache/pypoetry/virtualenvs/js-sdk-6NoanB64-py3.8/src/js-ng/jumpscale/entry_points/jsng.py", line 22, in run
    sys.exit(embed(globals(), locals(), configure=ptconfig, history_filename=HISTORY_FILENAME))
    │   │    │                                    │                          └ '/home/waleed/.jsng/history.txt'
    │   │    │                                    └ <function ptconfig at 0x7fc5f50b73a0>
    │   │    └ <function embed at 0x7fc5f4f57f70>
    │   └ <built-in function exit>
    └ <module 'sys' (built-in)>

  File "/home/waleed/.cache/pypoetry/virtualenvs/js-sdk-6NoanB64-py3.8/lib/python3.8/site-packages/ptpython/repl.py", line 350, in embed
    repl.run()
    │    └ <function PythonRepl.run at 0x7fc5f4f57af0>
    └ <ptpython.repl.PythonRepl object at 0x7fc5f64507f0>
  File "/home/waleed/.cache/pypoetry/virtualenvs/js-sdk-6NoanB64-py3.8/lib/python3.8/site-packages/ptpython/repl.py", line 74, in run
    self._process_text(text)
    │    │             └ 'j.servers.threebot.default.packages.install_all()'
    │    └ <function PythonRepl._process_text at 0x7fc5f4f57b80>
    └ <ptpython.repl.PythonRepl object at 0x7fc5f64507f0>
[...]

PR #159, opened 2020-07-01
change install_all method from being public

RafyAmgadBenjamin:
This is related to issue https://github.com/threefoldtech/js-sdk/issues/126
I have prevented the method from being public


waleedhammam:
@ahelsayd should approve

PR #160, opened 2020-07-01
add kraken client

alichaddad:


jumpscale/clients/kraken/kraken.py
```diff
@@ -0,0 +1,41 @@
...
+        super().__init__(*args,)
+        self._session = requests.Session()
+
+    def _do_request(self, url, ex):
+        res = self._session.get(url).json()
+        if res["error"]:
+            raise ex("\n".join(res["error"]))
+        return res
+
+    def get_pair_price(self, pair):
```
xmonader, line 23: can we make the default XLMUSD

Issue #166, opened 2020-07-02
Allow changing the chatflow title


xmonader:
Allow changing the title of a chatflow maybe by setting a class variable on the chatflow and if not let's leave the current default "Zero Chat Bot" as is

AhmedSa-mir:
Changed title here: https://github.com/threefoldtech/js-sdk/commit/9122fdcf0ed1f0ec511414cb73d75108d1b41871

xmonader:
```pythonn

class TheChatflow(Chatflow):
    TITLE = ....
```
Please make it similar to this

Dina-Abd-Elrahman:
- verified.

PR #185, opened 2020-07-05
Added chatflow title property

AhmedSa-mir:
Fixing https://github.com/threefoldtech/js-sdk/issues/166

### DIFF
--- a/jumpscale/packages/chatflows/actors/chatbot.py
+++ b/jumpscale/packages/chatflows/actors/chatbot.py
@@ -21,10 +21,10 @@ class ChatFlows(BaseActor):
         chatflow = package.get(chat)
         if not chatflow:
             raise j.exceptions.Value(f"Chat {chat} not found")
-        
+
         obj = chatflow()
         self.sessions[obj.session_id] = obj
-        return {"sessionId": obj.session_id}
+        return {"sessionId": obj.session_id, "title": obj.title}
 
     @actor_method
     def fetch(self, session_id: str, restore: bool = False) -> dict:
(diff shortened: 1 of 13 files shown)


jumpscale/packages/chatflows/actors/chatbot.py
```diff
@@ -21,10 +21,10 @@ def new(self, package: str, chat: str, client_ip: str, query_params: dict = None
         chatflow = package.get(chat)
         if not chatflow:
             raise j.exceptions.Value(f"Chat {chat} not found")
-        
+
         obj = chatflow()
         self.sessions[obj.session_id] = obj
-        return {"sessionId": obj.session_id}
+        return {"sessionId": obj.session_id, "title": obj.title}
```
xmonader, line 27: can we make sure we always send one incase of obj.title is None ?
e.g `"title": "Zero Chat bot" if obj.title is None else obj.title` ?

jumpscale/packages/chatflows/frontend/App.vue
```diff
@@ -105,6 +105,7 @@
         end: false,
         menu: false,
         chat: CHAT,
+        title: null,
```
xmonader, line 108: Can we have a default here instead of null?

jumpscale/packages/chatflows/actors/chatbot.py
AhmedSa-mir: We have a default title in the base class

jumpscale/packages/chatflows/frontend/App.vue
AhmedSa-mir: We have a default title in the base class
https://github.com/threefoldtech/js-sdk/blob/6f0932e4a1d4a9f74df611c613a47e2515ffcd46/jumpscale/sals/chatflows/chatflows.py#L83

Issue #177, opened 2020-07-05
chat has ended shows up if you don't press next in the last step when you come back to the chatflow


xmonader:
chat has ended shows up if you don't press next in the last step when you come back to the chatflow, because the session still exists and the last step "chat has ended" is still to be show

![image](https://user-images.githubusercontent.com/64129/86529103-2bffe900-beae-11ea-88ec-995cad529cf8.png)

grimpy:
@xmonader any suggestion how to solve this? Cause the chatflow engine does not know when the last step is the last step. unless we enforce the chatflows to indicate this somehow.

xmonader:
Hmm I think the chatflow_step decorator we can have last=True maybe and if we are that point we invalidate the session?

Dina-Abd-Elrahman:
- verified.

PR #199, opened 2020-07-07
Chatflows session invalidation on server restart

AhmedSa-mir:
See https://github.com/threefoldtech/js-sdk/issues/177

- Session is invalidated on client side on server restart instead of showing "Chat ended"
- Now when you refresh on the last step, it ends the chatflow.

### DIFF
--- a/jumpscale/packages/chatflows/actors/chatbot.py
+++ b/jumpscale/packages/chatflows/actors/chatbot.py
@@ -41,12 +41,7 @@ class ChatFlows(BaseActor):
 
     @actor_method
     def validate(self, session_id: str) -> dict:
-        print("AAAAAAAAAAAAA", session_id)
-        chatflow = self.sessions.get(session_id)
-        valid = True
-        if not chatflow:
-            valid = False
-        return {"valid": valid}
+        return {"valid": session_id in self.sessions}
 
     @actor_method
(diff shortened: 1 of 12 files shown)


jumpscale/packages/chatflows/actors/chatbot.py
```diff
@@ -39,6 +39,15 @@ def fetch(self, session_id: str, restore: bool = False) -> dict:
 
         return result
 
+    @actor_method
+    def validate(self, session_id: str) -> dict:
+        print("AAAAAAAAAAAAA", session_id)
```
ahelsayd, line 44: remove this

jumpscale/packages/chatflows/actors/chatbot.py
```diff
@@ -39,6 +39,15 @@ def fetch(self, session_id: str, restore: bool = False) -> dict:
...
+        chatflow = self.sessions.get(session_id)
+        valid = True
+        if not chatflow:
+            valid = False
+        return {"valid": valid}
+
```
ahelsayd, line 50: why not just ?
```python
return {"valid": session_id in self.sessions}
```

AhmedSa-mir: nice thnx :)

PR #217, opened 2020-07-12
Development stellar helpers

waleedhammam:
- Change place sell order to manage sell order to have the ability to delete current order
- add a method to get the current placed orders


jumpscale/clients/stellar/stellar.py
```diff
@@ -835,22 +836,24 @@ def get_asset(self, code="TFT", issuer=None) -> stellar_sdk.Asset:
             asset_issuer = _NETWORK_KNOWN_TRUSTS[network].get(code, None)
             return Asset(code, asset_issuer)
 
-    def place_sell_order(
+    def manage_sell_order(
```
xmonader, line 839: Let's create more helpers 
e.g (cancel_sell_order or delete_sell_order), better imo to pass different amount as flag for deletion

waleedhammam: Done and testd, place_sell_order and cancel_sell_order

jumpscale/clients/stellar/stellar.py
```diff
@@ -892,6 +911,8 @@ def manage_sell_order(
                 raise RuntimeError(f"couldn't sumbit transaction, probably unfunded") from e
```
xmonader, line 911: typo should be submit instead of sumbit

waleedhammam: actually, this whole message should be refactored, cause it only shows underfunded transaction, we can occur more errors like offer not found or whatever

Issue #275, opened 2020-07-21
codeserver package seems to install incorrectly


xmonader:
# what happens

- start 3bot
- go to /admin
- codeserver
- install package

I see 

![image](https://user-images.githubusercontent.com/64129/88053216-b8233900-cb5b-11ea-8b76-06f1b7f10627.png)

instead of codeserver ...

dmahmouali:
Can't reproduce location error, but find anther error related to load static files

![photo_2020-07-21_14-49-22](https://user-images.githubusercontent.com/36021484/88056797-68e00700-cb61-11ea-8d58-414c753a701c.jpg)

Dina-Abd-Elrahman:
- verified.

PR #289, opened 2020-07-22
Configuring buffering in nginx location conf

AhmedSa-mir:
See https://github.com/threefoldtech/js-sdk/issues/275#issuecomment-661840421


jumpscale/sals/nginx/templates/nginx.conf
```diff
@@ -32,7 +32,7 @@ http {
     proxy_connect_timeout       1d;
     proxy_send_timeout          1d;
     proxy_read_timeout          1d;
-
+    proxy_buffering off;
```
abom, line 35: This can be moved on location level, we can add a new options for buffering in `Location`.

abom: This will avoid setting it globally in the main nginx.conf

abom: Also, this problem cannot be resolved by setting correct `proxy_buffer_size` and `proxy_buffers` (also in the location level)?

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size

jumpscale/packages/codeserver/package.toml
```diff
@@ -9,3 +9,4 @@ port = 8080
 path_url = "/codeserver"
 path_dest = "/"
 websocket = true
+increase_buffers_size = true
```
abom, line 12: I think we can just add the full options, not increase buffers size.
This won't fit all cases.

jumpscale/servers/threebot/threebot.py
```diff
@@ -142,6 +142,7 @@ def apply(self):
                         loc.port = location.get("port", PORTS.HTTP)
                         loc.path_dest = location.get("path_dest", "")
                         loc.websocket = location.get("websocket", False)
+                        loc.increase_buffers_size = location.get("increase_buffers_size", False)
```
abom, line 145: If you add a property, it should be added to `Location class too, this way, it won't be stored.

jumpscale/packages/notebooks/package.toml
```diff
@@ -9,3 +9,5 @@ port = 8888
 path_url = "/notebooks"
 path_dest = "/notebooks"
 websocket = true
+nginx_proxy_buffers = "8 256k"
```
abom, line 12: I think `proxy_buffers` is enough? (without nginx)
Also, add buffering enabled field too.

PR #323, opened 2020-07-23
Development marketplace improvements

m-motawea:


jumpscale/packages/marketplace/chats/threebot.py
```diff
@@ -7,18 +7,20 @@
...
+from jumpscale.data.nacl.jsnacl import NACL
 
 
 class ThreebotDeploy(MarketPlaceChatflow):
     SOLUTION_TYPE = SolutionType.Threebot
     steps = [
         "welcome",
-        "solution_name",
+        "set_solution_name",
+        "set_backup_password",
```
xmonader, line 18: let's use generic password instead of backup password

Issue #354, opened 2020-07-28
Enhance admin dashboard to render readable data in workload details [development_enhance_dashboard]


dmahmouali:
![image](https://user-images.githubusercontent.com/36021484/88669493-40698700-d0e4-11ea-91b9-c922e586d41f.png)

dmahmouali:
Used new json viewer
![image](https://user-images.githubusercontent.com/36021484/88814780-d839a480-d1ba-11ea-81db-c695e13a0f3b.png)
and added new details view
![image](https://user-images.githubusercontent.com/36021484/88814877-f43d4600-d1ba-11ea-9841-79b00c95b400.png)

Dina-Abd-Elrahman:
- verified.

PR #360, opened 2020-07-29
update solution details info view

dmahmouali:
issue : https://github.com/threefoldtech/js-sdk/issues/354
Added new details view
image
and used new json viewer
image

### DIFF
--- a/jumpscale/packages/admin/frontend/components/solutions/Info.vue
+++ b/jumpscale/packages/admin/frontend/components/solutions/Info.vue
@@ -2,10 +2,42 @@
   <div>
     <base-dialog :title="title" v-model="dialog" :loading="loading">
       <template #default>
-        <code-area mode="python" :content="json"></code-area>
+        <v-list-item v-for="(item, key)  in json" :key="key">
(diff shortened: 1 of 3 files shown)


jumpscale/packages/admin/frontend/components/solutions/Info.vue
```diff
@@ -2,10 +2,49 @@
...
+          <v-tab :key="'moredetails'">More details</v-tab>
+
+          <v-tab-item :key="title">
+            <v-simple-table>
+              <template v-slot:default>
+                <tbody>
+                  <tr v-for="(item, key)  in json" :key="key">
+                    <th>{{ key }}</th>
+                    <td
+                      v-if="key === 'Node ids' || key === 'wids' || key === 'Active workload ids'"
```
AhmedSa-mir, line 16: Keys should be defined as an array and check `v-if="array.includes(key)"`, no?
so that we can add other keys later.

dmahmouali: changed, thanks Ahmed :+1:

PR #378, opened 2020-08-03
 fix when delete corrupted package

BolaNasr:


jumpscale/servers/threebot/threebot.py
```diff
@@ -199,15 +199,18 @@ class Package:
     def __init__(self, path, default_domain, default_email, giturl=""):
         self.path = path
         self.giturl = giturl
-        self.config = self.load_config()
-        self.name = self.config["name"]
+        self._config = None
+        self.name = j.sals.fs.path_parts(f"{path}")[-1]
```
xmonader, line 203: why not j.sal.fs.basename? and if it's the same variable and no need to format it should be `j.sals.fs.path_parts(path)`

ahsayde, line 203: better to do it like 

```python
self.name = j.sals.fs.basename(path)
```

jumpscale/servers/threebot/threebot.py
```diff
@@ -199,15 +199,18 @@ class Package:
...
         self._module = None
         self.default_domain = default_domain
         self.default_email = default_email
 
     def load_config(self):
-        return toml.load(j.sals.fs.join_paths(self.path, "package.toml"))
+        path = j.sals.fs.join_paths(self.path, "package.toml")
+        if j.sals.fs.exists(path):
+            return toml.load(path)
+        return None
```
ahsayde, line 213: keep it as it is, we need it to fail if the config is not exists maybe you can just check if the package.toml exists and raise an error if it is not

jumpscale/servers/threebot/threebot.py
```diff
@@ -416,14 +427,15 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
         package = self.get(package_name)
         if not package:
             raise j.exceptions.NotFound(f"{package_name} package not found")
 
         # remove bottle servers
-        for bottle_server in package.bottle_servers:
-            self.threebot.rack.remove(f"{package.name}_{bottle_server['name']}")
+        for bottle_server in self.threebot.rack._servers:
+            if package_name in bottle_server:
+                self.threebot.rack.remove(f"{bottle_server}")
```
waleedhammam, line 437: I think it should be `f"{package.name}_{bottle_server['name']}"` to grep the correct server

jumpscale/servers/threebot/threebot.py
```diff
@@ -416,14 +427,15 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
+                break
```
waleedhammam, line 438: Also the package could have more than server maybe we should let it continue the loop

jumpscale/servers/threebot/threebot.py
```diff
@@ -225,6 +228,14 @@ def module(self):
     def base_url(self):
         return j.sals.fs.join_paths("/", self.name)
 
+    @property
+    def config(self):
+        if not self._config:
+            self._config = self.load_config()
+            if self._config:
+                self.name = self._config["name"]
```
ahsayde, line 236: why you reset the name here ?
it should be like 
```python
if self._config is None:
    self._config = self.load_config()
return self._config
```

jumpscale/servers/threebot/threebot.py
```diff
@@ -416,14 +427,15 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
             raise j.exceptions.Value("cannot delete default packages")
-
         package = self.get(package_name)
         if not package:
             raise j.exceptions.NotFound(f"{package_name} package not found")
 
         # remove bottle servers
-        for bottle_server in package.bottle_servers:
-            self.threebot.rack.remove(f"{package.name}_{bottle_server['name']}")
+        for bottle_server in self.threebot.rack._servers:
```
ahsayde, line 435: better add *servers* property in the rack

jumpscale/servers/threebot/threebot.py
```diff
@@ -416,14 +427,15 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
+            if package_name in bottle_server:
```
ahsayde, line 436: this check is not enough 
for example you have two packages *farm* & *farm_manager* and you want to remove package *farm*
with this condition the servers of package *farm_manager* will be deleted to because farm in farm_manager
you need to do something like
```python
if bottle_server.startswith(f"{package_name}_"):
```

jumpscale/servers/threebot/threebot.py
ahsayde, line 438: why do u break here ?
we want to remove all the bottle servers of the package

ahsayde:
also you need to handle unregistering package's actor in case threebot is running

jumpscale/servers/threebot/threebot.py
BolaNasr: I used ```j.sals.fs.path_parts(path)``` because if we add  ```/ ``` 
 in the end of the path like 
``` /sandbox/code/github/threefoldtech/js-sdk/jumpscale/packages/marketplace/ ```path_parts function igonre it and return basename but basename function will return / not basename.

Comment (BolaNasr) [reply jumpscale/servers/threebot/threebot.py:203]: I used ```j.sals.fs.path_parts(path)``` because if we add  ```/ ``` 
 in the end of the path like 
``` /sandbox/code/github/threefoldtech/js-sdk/jumpscale/packages/marketplace/ ```path_parts function igonre it and return basename but basename function will return / not basename.

File: jumpscale/servers/threebot/threebot.py (hunk 5 above)
Comment (BolaNasr) [reply jumpscale/servers/threebot/threebot.py:236]: because before name is in init and name is get from config , so when config is exist and loaded get name otherwise not change it .

File: jumpscale/servers/threebot/threebot.py (hunk 2 above)
Comment (BolaNasr) [reply jumpscale/servers/threebot/threebot.py:213]: but it's a bug because when raise error , i can't delete it .
the issue is how to delete corrupted package , so what if package.toml doesn't exist and this package is added before , should i can delete it

File: jumpscale/servers/threebot/threebot.py (hunk 1 above)
Comment (ahsayde) [reply jumpscale/servers/threebot/threebot.py:203]: you can do 
```python
j.sals.fs.basename(path.strip('/'))
```

jumpscale/servers/threebot/threebot.py
ahsayde: we will not call load_config in the delete method so it will not be a problem

jumpscale/servers/threebot/threebot.py
ahsayde: we already now get the name using this line
```j.sals.fs.basename(path.rstrip('/'))```

File: jumpscale/servers/threebot/threebot.py [hunk 8]
```diff
@@ -416,20 +422,20 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
+                self.threebot.rack.remove(f"{bottle_server}")
+
 
         if self.threebot.started:
             # unregister gedis actors
-            if package.actors_dir:
-                for actor in package.actors:
-                    self.threebot.gedis._system_actor.unregister_actor(actor["name"])
+            for actor in self.threebot.gedis._loaded_actors.keys():
+                self.threebot.gedis._system_actor.unregister_actor(actor)
```
ahsayde, line 438: where is the condition ? all the actors will be unregistered

jumpscale/servers/threebot/threebot.py
```diff
@@ -416,20 +422,20 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
         package = self.get(package_name)
         if not package:
             raise j.exceptions.NotFound(f"{package_name} package not found")
 
         # remove bottle servers
-        for bottle_server in package.bottle_servers:
-            self.threebot.rack.remove(f"{package.name}_{bottle_server['name']}")
+        for bottle_server in self.threebot.rack._servers:
+            if bottle_server.startswith(f"{package_name}_"):
+                self.threebot.rack.remove(f"{bottle_server}")
```
ahsayde, line 432: why u format the variable here ?

PR #395, opened 2020-08-06
Peertube

ashraffouda:


jumpscale/packages/marketplace/frontend/components/solutions/Solutions.vue
```diff
@@ -131,6 +131,12 @@ module.exports = {
           description:
             "Deploy a wiki, blog, website and access it via an IP and a public domain"
         },
+        peertube: {
```
m-motawea, line 134: I see the chatflow is added to marketplace ui. but there is no chatflow for it in the marketplace

ashraffouda: https://github.com/threefoldtech/js-sdk/pull/395/files#diff-0233c27fe1c7e27faa5ea76b038381f0

Issue #339, opened 2020-07-26
threebot server load test


alichaddad:
There have been some problems with Threebot server performance, need to perform load and test and determine the root issue.

RafyAmgadBenjamin:
I have used Locust python package to do Load test
#### Testing using NGINX
- with 200 users and hatch rate (number of users to spawn per second) is 10 
![200User_10H](https://user-images.githubusercontent.com/46320202/88817787-39af4280-d1be-11ea-873e-101622cd2810.jpg)
- with 2000 users and hatch rate is 20 
![2000User_20H_before_update_nginx](https://user-images.githubusercontent.com/46320202/88817834-4af84f00-d1be-11ea-8ae0-41d9b4d8ca3f.jpg)

then we update Nginx configuration 
![Screenshot from 2020-07-29 16-50-22](https://user-images.githubusercontent.com/46320202/88817865-551a4d80-d1be-11ea-8840-a2662049cb55.png)
- with 2000 users and hatch rate is 20 
![image](https://user-images.githubusercontent.com/46320202/88817918-64999680-d1be-11ea-8a29-a1c36c3c97ee.png)

RafyAmgadBenjamin:
##### commits 
https://github.com/threefoldtech/js-ng/commit/a7327f311471849c91f905290d3d0892e542e6e3
https://github.com/threefoldtech/js-ng/commit/2ca6104e3ab368dc64aa3fdc569ff0620dbed5cb

RafyAmgadBenjamin:
I have added the code in js-sdk, update the docs and make the path as a parameter can be passed to the bash script
#### Commit
https://github.com/threefoldtech/js-sdk/commit/5d8f01a4ff4950952610927bf5074ac5335af80a

RafyAmgadBenjamin:
Make the port can be passed as an option 
PR 
https://github.com/threefoldtech/js-sdk/pull/400

PR #400, opened 2020-08-06
Development load testing

RafyAmgadBenjamin:
#### Issue
https://github.com/threefoldtech/js-sdk/issues/339

- Create load testing
- Add bash script to run the load testing
- Add docs

### DIFF
--- a/tests/loadtesting/README.md
+++ b/tests/loadtesting/README.md
@@ -48,6 +48,8 @@ for more information about `3bot` starting [check](https://github.com/threefoldt
 - `r` The rate per second in which users are spawned (obligatory)
 - `h` Host to load test (obligatory)
 - `p` Path for python script which uses locust in load testing (obligatory)
+- `n` Number of port used in the request, default is 8000 (optional)
+ 
 
 ##### Examples
 - Help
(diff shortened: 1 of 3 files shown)


tests/loadtesting/README.md
```diff
@@ -0,0 +1,75 @@
+### Load testing
+
+##### Introduction
+we have applied `Load test` to test the performance of our admin endpoints served through [threebot server](https://github.com/threefoldtech/js-sdk), we have used `Locust` to  assess the performance.
+
+#### Technology Used
+- `Threebot server` which is Jumpscale is a cloud automation product and a webplatform to develop web application.
```
abom, line 7: Can reference threebot docs directly instead?

tests/loadtesting/README.md
```diff
@@ -0,0 +1,75 @@
...
+To see available options, run:
+```bash
+$ locust --help
+```
+for more information about the installation [check](https://docs.locust.io/en/stable/installation.html)
+
+#### Run load test
+1. Run `3bot`
+
+- open `jsng` shell then
```
abom, line 37: This can be removed, as we mention the quick start doc for threebot.

tests/loadtesting/admin.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class QuickstartUser(HttpUser):
+    # Our class defines a wait_time function that will make the simulated users
+    #  wait between 5 and 9 seconds after each task is executed
+    wait_time = between(5, 9)
+
+    @task
+    def index_page(self):
+        self.client.get("https://localhost/admin/#/", verify=False)
```
abom, line 13: No need to hard-code the full URL here, just use the endpoint without `https://localhost' and it should use the host you provided?

RafyAmgadBenjamin: This will work with the first two enpoints and they don't require a port but in the last endpoint, it didn't work due to the presence of the port number

xmonader:
I merged let's improve in the next iterationn

PR #403, opened 2020-08-06
avoid having circular dependency between explorer client and identity

xmonader:


jumpscale/core/identity/__init__.py
```diff
@@ -101,10 +102,19 @@ def tid(self):
 
     def register(self, host=None):
         self.verify_configuration()
-        try:
-            user = self.explorer.users.get(name=self.tname)
-        except NotFound:
+
+        # check if we are not already registered with the configured identity
+        resp = requests.get(f"{self.explorer_url}/users?name={self.tname}")
+        if resp.status_code == 404:
```
waleedhammam, line 108: I think no need for that cause it will always return 200 even if user is not found

Issue #391, opened 2020-08-05
Create automated test for payment


AhmedHanafy725:
All payment now on `testnet` are done on capacity pools.
So it supposes that the test to create a pool and pay for it.

AhmedHanafy725:
#440

PR #451, opened 2020-08-11
add payment tests

AhmedHanafy725:
#391

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ integrationtests:
 	pytest tests -sv -m "integration"
 
 unittests:
-	pytest tests -sv -m "not integration"
+	pytest tests -sv -m "unittests"
 
 docs:
 	pdoc3 jumpscale --html --output-dir docs/api --overwrite
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,13 @@
 .PHONY: tests docs
 
 tests:
-	pytest tests
+	pytest tests -sv
+
+integrationtests:
+	pytest tests -sv -m "integration"
+
+unittests:
+	pytest tests -sv -m "not integration"
 
 docs:
 	pdoc3 jumpscale --html --output-dir docs/api --overwrite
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,7 +1,13 @@
...
 
 tests:
-	pytest tests
+	pytest tests -sv
+
+integrationtests:
+	pytest tests -sv -m "integration"
+
+unittests:
+	pytest tests -sv -m "not integration"
```
abom, line 10: I think "unittests" mark would be better instead of integration/not integration?

tests/integration/test_payment.py
```diff
@@ -0,0 +1,184 @@
+import os
+import string
+import time
+
+import pytest
+
+from jumpscale.loader import j
+
+WALLET_NAME = os.environ.get("WALLET_NAME")
```
abom, line 9: who should set this?

tests/integration/test_payment.py
```diff
@@ -0,0 +1,184 @@
...
+    if WALLET_NAME and WALLET_SECRET:
+        wallet = j.clients.stellar.get(WALLET_NAME, network="TEST", secret=WALLET_SECRET)
+        return wallet
+    else:
+        raise ValueError("Please provide add Values to the environment variables WALLET_NAME and WALLET_SECRET")
+
+
+def create_new_wallet():
+    wallet_name = j.data.idgenerator.nfromchoices(10, string.ascii_letters)
+    wallet = j.clients.stellar.new(wallet_name, network="TEST")
```
abom, line 29: a client is created here and other config objects, do we save them? and do we cleanup later?

AhmedHanafy725: No, it's not saved and they will die once the test is finished.

tests/integration/test_payment.py
AhmedHanafy725: Anyone who wants to run the test, and it will raise an error if it's not set.

PR #455, opened 2020-08-11
simplify publisher questions

waleedhammam:


jumpscale/packages/tfgrid_solutions/chats/publisher.py
```diff
@@ -127,20 +116,15 @@ def domain_select(self):
...
+        self.domain = f"{self.sub_domain}.{self.domain}"
         self.envars["DOMAIN"] = self.domain
-
         self.addresses = []
         for ns in self.gateway.dns_nameserver:
             self.addresses.append(j.sals.nettools.get_host_by_name(ns))
 
         self.secret = f"{j.core.identity.me.tid}:{uuid.uuid4().hex}"
 
-    @chatflow_step(title="Global IPv6 Address")
```
m-motawea, line 140: the IPv6 was requested before. are you sure about removing it here?

PR #462, opened 2020-08-12
fix farm filtering

m-motawea:


jumpscale/sals/reservation_chatflow/deployer.py
```diff
@@ -288,15 +288,19 @@ def check_farm_capacity(self, farm_name, currencies=[], sru=None, cru=None, mru=
         available_sru = 0
         available_mru = 0
         available_hru = 0
+        running_nodes = 0
         for node in farm_nodes:
             if "FreeTFT" in currencies and not node.free_to_use:
                 continue
             if not j.sals.zos.nodes_finder.filter_is_up(node):
                 continue
+            running_nodes += 1
```
zaibon, line 297: you need to actually check that the node is online before you can increment this counter.
Node are supposed to report they status every 10 minutes, so any node that has a `updated` field with a value of more then 10 minutes ago is considered offline.

m-motawea: isn't that what the `j.sals.zos.nodes_finder.filter_is_up(node)` does?

PR #478, opened 2020-08-12
Development capacitypool cryptpad

BolaNasr:


jumpscale/packages/tfgrid_solutions/chats/cryptpad_deploy.py
```diff
@@ -49,7 +50,7 @@ def cryptpad_name(self):
                     break
                 valid = True
 
-    @chatflow_step(title="Container resources")
+    @chatflow_step(title="")
```
waleedhammam, line 53: it's not a step actually, we can merge it with `volume_details` method and give it a title `Solution Size` or so

jumpscale/packages/tfgrid_solutions/chats/cryptpad_deploy.py
```diff
@@ -81,6 +82,9 @@ def select_pool(self):
     @chatflow_step(title="Network")
     def cryptpad_network(self):
         self.network_view = deployer.select_network(self)
+
+    @chatflow_step(title="")
+    def container_node_id(self):
```
waleedhammam, line 87: we can merge this too with container ip with domain, under title `Choose your domain`

Issue #476, opened 2020-08-12
workload detail vue should show a proper UI for workload.info field


zaibon:
Right now we just print the content as a weird dump:
![2020-08-12_1937x216_scrot](https://user-images.githubusercontent.com/303164/90021682-e35bfc80-dcb1-11ea-85dd-aeb1cbb0e3dd.png)

dmahmouali:
Added three filters on workload data:
1. `KeysIgnored` : to ignore keys in workload tab view
2. `KeysWithTypeList` : to view keys with type list in `v-chip` tag
3. `KeysWithTypeDict` : to view keys with type dict in `v-chip` tag with pattern `key : value`

![image](https://user-images.githubusercontent.com/36021484/90395080-293f0900-e094-11ea-820b-e1437bca281a.png)
![image](https://user-images.githubusercontent.com/36021484/90395161-55f32080-e094-11ea-9f76-5ac4acabcbfa.png)

PR #504, opened 2020-08-17
Development capacitypool ui improve

dmahmouali:
issue: https://github.com/threefoldtech/js-sdk/issues/476

Added three filters on workload data:
1. `KeysIgnored` : to ignore keys in workload tab view
2. `KeysWithTypeList` : to view keys with type list in `v-chip` tag
3. `KeysWithTypeDict` : to view keys with type dict in `v-chip` tag with pattern `key : value`


abom:
I asked @dmahmouali to make it a single component and re-use it, as we have nearly the same code duplicated in another view.

Issue #506, opened 2020-08-17
Improve flow for new app store chatflows


ranatrk:
complimentary to: https://github.com/threefoldtech/home/issues/878

_**Proposal:**_
- Scenarios to be handled for the service payments:
  - Wrong payment(wrong amount or wrong payment id in memo text of transaction)
  - Should wait on payment that it was successful
  - Options for handling service fees
    - **Option 1**: Need to set the fees for the use of the service which will be a one time payment when they first access. The users that payed will be kept track of on the machine (similar to trader and polls voting mechanism) and the payments for the pool deployments are done from user to explorer directly no double transfer is done.
      <img src="https://user-images.githubusercontent.com/17128393/90388687-80d77780-e088-11ea-914f-d8af53ddbc9c.png" width="500" height="500" />
    - **Option 1-b**: Need to set the fees for the use of the service which will be payed everytime to the app store's wallet along with the pool creation fees. Wont need to keep track of who payed these fees as it will be obligatory with every deployment process. In case the deployment failed for any reason, the fees can be rfunded to the user.

    - **Option 2**: Fees are added to the cost of the deployment, and the tokens are transferred to the app store's wallet, and the app store deducts the fees and proceeds with the payment for the deployment (Disadvantage: not sure if multiple transactions can be done simultaneously if we have multiple deployments from different users at the same time using only 1 app store wallet) 
      <img src="https://user-images.githubusercontent.com/17128393/90388547-4241bd00-e088-11ea-94ee-a2369c980f14.png" width="500" height="500" />

   - **Option 3**: Three large pools are created by the app store machine and tokens are transferred to the app's store wallet by the user. The app store will own and be responsible in managing pools and deployments. (Disadvantage: if the pool resources finish, all deployments on that pool are cancelled)

**_To be done_**

- Deployment scenarios to be handled:
  - when someone pays for a pool and deploys network,but by the time they create a container the resources finished and the deployment fails, then their money is lost. The failed deployments can be saved for that user where they can list them in the UI and have an option to deploy later or cancel that attempt in the first place. This will ensure the pool they paid for is not lost but doesn't guarantee that resources will free up.

BolaNasr:
I agree with **Option 1**

rkhamis:
I think the first suggestion has less edge cases and less complicated

ranatrk:
After discussing with @waleedhammam @ashraffouda ,and @rkhamis : **Option 1-b** will be approached. 
[...]

PR #508, opened 2020-08-17
Add fees payment to app store chatflows

ranatrk:
issue: https://github.com/threefoldtech/js-sdk/issues/506
with publisher chatflows used as an example


jumpscale/sals/marketplace/chatflow.py
```diff
@@ -169,9 +191,78 @@ def solution_name(self):
     @chatflow_step(title="Payment currency")
     def payment_currency(self):
         self.currency = self.single_choice(
-            "Please select the currency you want to pay with.", ["FreeTFT", "TFT", "TFTA"], required=True
+            "Please select the currency you want to pay with.", ["FreeTFT", "TFT", "TFTA"], required=True, default="TFT"
+        )
+
+    def _refund(self, transaction, effects):
+        currency = self.currency or "TFT"
+        if effects.amount < 0:
```
m-motawea, line 199: shouldn't we check if amount <= 0.1 the transaction fees?

waleedhammam: indeed you are partially right, in TFT there's 0.1 transaction fees., TFTA&FreeTFT no but not included in the effects

so in the refund if he paid with TFT, we should deduce 0.1 from the refunded amount as a transaction fees

Issue #364, opened 2020-07-30
Add health check for stellar network / tokenservices


xmonader:
Make sure there's an indicator in the admin dashboard / marketplace to show the status of stellar network (based on the active identity) and the health of the tokenservices.

xmonader:
@waleedhammam please add the links for the checks

waleedhammam:
website: https://testnet.threefold.io/
api: https://testnet.threefold.io/threefoldfoundation/transactionfunding_service/fund_transaction
traceback in case of transaction fail

```
  resp = j.tools.http.post(self._get_url("FUND"), json={"args": data})
           │ │               │    │                               └ {'transaction': 'AAAAADrJoEr0EGzHGBuoKiFCaNbibRmYEyMuxrkSB8rVrZEbAAAAZAAABAQAAAALAAAAAQAAAAAAAAAAAAAAAF8sBCgAAAABAAAABXAtMTA2...
           │ │               │    └ <function Stellar._get_url at 0x7f8ae55b9f70>
           │ │               └ Stellar(_Base__instance_name='wallet', _Base__parent=None, __address='GA5MTICK6QIGZRYYDOUCUIKCNDLOE3IZTAJSGLWGXEJAPSWVVWIRWXA...
           │ └ <property object at 0x7f8aeb963ea0>
           └ <jumpscale.loader.J object at 0x7f8aeb95e550>

  File "/home/waleed/.cache/pypoetry/virtualenvs/js-sdk-6NoanB64-py3.8/lib/python3.8/site-packages/requests/api.py", line 119, in post
    return request('post', url, data=data, json=json, **kwargs)
           │               │    │    │     │    │       └ {}
           │               │    │    │     │    └ {'args': {'transaction': 'AAAAADrJoEr0EGzHGBuoKiFCaNbibRmYEyMuxrkSB8rVrZEbAAAAZAAABAQAAAALAAAAAQAAAAAAAAAAAAAAAF8sBCgAAAABAAA...
           │               │    │    │     └ {'args': {'transaction': 'AAAAADrJoEr0EGzHGBuoKiFCaNbibRmYEyMuxrkSB8rVrZEbAAAAZAAABAQAAAALAAAAAQAAAAAAAAAAAAAAAF8sBCgAAAABAAA...
           │               │    │    └ None
           │               │    └ None
           │               └ 'https://testnet.threefold.io/threefoldfoundation/transactionfunding_service/fund_transaction'
[...]

PR #510, opened 2020-08-18
Adding dashboard health checks

AhmedSa-mir:
See https://github.com/threefoldtech/js-sdk/issues/364
Now stellar check only is available

stellar

### DIFF
--- a/jumpscale/packages/admin/actors/health.py
+++ b/jumpscale/packages/admin/actors/health.py
@@ -34,5 +34,17 @@ class Health(BaseActor):
     def get_running_processes(self) -> str:
         return j.data.serializers.json.dumps({"data": j.sals.process.get_processes_info()})
 
+    @actor_method
+    def get_health_checks(self) -> str:
+        checks = {"stellar": True}
+        try:
+            j.tools.http.get("https://horizon-testnet.stellar.org")
+            j.tools.http.get(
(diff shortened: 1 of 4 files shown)


jumpscale/packages/admin/actors/health.py
```diff
@@ -34,5 +34,17 @@ def get_memory_usage(self) -> str:
     def get_running_processes(self) -> str:
         return j.data.serializers.json.dumps({"data": j.sals.process.get_processes_info()})
 
+    @actor_method
+    def get_health_checks(self) -> str:
+        checks = {"stellar": True}
+        try:
+            j.tools.http.get("https://horizon-testnet.stellar.org")
+            j.tools.http.get(
+                "https://testnet.threefold.io/threefoldfoundation/transactionfunding_service/fund_transaction"
```
abom, line 43: why this is hard-coded here? and what if we are using the mainnet?

abom: Maybe we can have different check according to every network, or show the current identity's network check only?

AhmedSa-mir: yes I'm just waiting for Waleed's approval about the urls.

AhmedSa-mir:
> there should be a healthcheck for tokenservices along with the stellar health check too

Added

Issue #143, opened 2020-06-30
Unable to delete a package for which the files don't exist anymore


grimpy:
```python
In [4]: j.servers.threebot.default.packages.delete('admin2')                                                                                      
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-4-a9e830357665> in <module>
----> 1 j.servers.threebot.default.packages.delete('admin2')

~/sandbox/code/github/threefoldtech/js-sdk/jumpscale/servers/threebot/threebot.py in delete(self, package_name)
    331             raise j.exceptions.Value("cannot delete default packages")
    332 
--> 333         package = self.get(package_name)
    334         if not package:
    335             raise j.exceptions.NotFound(f"{package_name} package not found")

~/sandbox/code/github/threefoldtech/js-sdk/jumpscale/servers/threebot/threebot.py in get(self, package_name)
    279         package_path = self.packages.get(package_name)
    280         if package_path:
--> 281             return Package(path=package_path, default_domain=self.threebot.domain, default_email=self.threebot.email)
    282 
    283     def get_packages(self):

~/sandbox/code/github/threefoldtech/js-sdk/jumpscale/servers/threebot/threebot.py in __init__(self, path, default_domain, default_email)
    164     def __init__(self, path, default_domain, default_email):
    165         self.path = path
--> 166         self.config = self.load_config()
    167         self.name = self.config["name"]
    168         self.nginx_config = NginxPackageConfig(self)

~/sandbox/code/github/threefoldtech/js-sdk/jumpscale/servers/threebot/threebot.py in load_config(self)
    172 
    173     def load_config(self):
--> 174         return toml.load(j.sals.fs.join_paths(self.path, "package.toml"))
    175 
    176     @property

~/sandbox/code/github/threefoldtech/js-ng/.venv/lib/python3.8/site-packages/toml/decoder.py in load(f, _dict, decoder)
    131 
    132     if _ispath(f):
--> 133         with io.open(_getpath(f), encoding='utf-8') as ffile:
    134             return loads(ffile.read(), _dict, decoder)
    135     elif isinstance(f, list):

FileNotFoundError: [Errno 2] No such file or directory: '/home/Jo/sandbox/code/github/threefoldtech/js-sdk/jumpscale/packages/admin2/package.toml'
```

grimpy:
We could just detect this and pop it out of the configured dict and show warning...
I worked around this by doing:
```python
[...]

PR #528, opened 2020-08-19
Development  delete corrupted package

RafyAmgadBenjamin:
### Description

Handle the deletion of corrupted package, prevent it from raising an exception, remove it from the configuration dictionary, and raise a warning in this case. 

### Related Issues
https://github.com/threefoldtech/js-sdk/issues/143

### DIFF
--- a/jumpscale/servers/threebot/threebot.py
+++ b/jumpscale/servers/threebot/threebot.py
@@ -441,8 +441,13 @@ class PackageManager(Base):
                     self.threebot.gedis._system_actor.unregister_actor(actor)
 
             # unload chats
-            if package.chats_dir:
-                self.threebot.chatbot.unload(package.chats_dir)
+            try:
+                if package.chats_dir:
+                    self.threebot.chatbot.unload(package.chats_dir)
+            except:
+                j.core.logging.error(
+                    f"Error during unloading the chats related to {package_name} package, maybe the package.toml file is not found."
+                )
 
             # reload nginx
             self.threebot.nginx.reload()
--- a/jumpscale/servers/threebot/threebot.py
+++ b/jumpscale/servers/threebot/threebot.py
@@ -445,7 +445,7 @@ class PackageManager(Base):
                 if package.chats_dir:
                     self.threebot.chatbot.unload(package.chats_dir)
             except:
-                j.core.logging.warning(
+                j.logger.warning(
                     f"Error during unloading the chats related to {package_name} package, maybe the package.toml file is not found."
                 )
 
--- a/jumpscale/servers/threebot/threebot.py
+++ b/jumpscale/servers/threebot/threebot.py
@@ -445,7 +445,7 @@ class PackageManager(Base):
                 if package.chats_dir:
                     self.threebot.chatbot.unload(package.chats_dir)
             except:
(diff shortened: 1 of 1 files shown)


BolaNasr (approved):
LGTM

jumpscale/servers/threebot/threebot.py
```diff
@@ -441,8 +441,13 @@ def delete(self, package_name):
                     self.threebot.gedis._system_actor.unregister_actor(actor)
 
             # unload chats
-            if package.chats_dir:
-                self.threebot.chatbot.unload(package.chats_dir)
+            try:
+                if package.chats_dir:
+                    self.threebot.chatbot.unload(package.chats_dir)
+            except:
+                j.core.logging.warning(
```
xmonader, line 448: we can use j.logger directly

jumpscale/servers/threebot/threebot.py
```diff
@@ -441,8 +441,13 @@ def delete(self, package_name):
...
+                    f"Error during unloading the chats related to {package_name} package, maybe the package.toml file is not found."
```
xmonader, line 449: is that the cause? how the package got registered without validating the package first? also can we improve the phrasing a bit e.g "couldn't unload chats of {package_name} in path ...."

also I think we need to improve the handling of the unnnloadingn itself https://github.com/threefoldtech/js-sdk/blob/741ff0ba820bd72c7113f3f8aff7ffeb51d10f3e/jumpscale/packages/chatflows/actors/chatbot.py#L100, and in the loggin message at least we don't lose the information of the current exception / stacktrace, best we create an alert

xmonader (changes requested):
Added some comments regarding the logging and not losing the error stacktraces

jumpscale/servers/threebot/threebot.py
RafyAmgadBenjamin: Done

jumpscale/servers/threebot/threebot.py
RafyAmgadBenjamin: - I want to clarify something, I am handling an error occurred after adding the package successfully, like deletion of package.toml file then delete the package,
- This part raises an exception so early as it tries to get `chats_dir` so it tries to read the package.toml.
https://github.com/threefoldtech/js-sdk/blob/b00ce331f02068ee45a4010345956ca73dce0cba/jumpscale/servers/threebot/threebot.py#L445-L446

RafyAmgadBenjamin: Regarding the logging part. so you want me to show a warning and log the message with the actual exception, is it right?

xmonader: Exceptions shouldn't be hidden or masked, if we except any exception we catch it (so we have access to the raised exception and its data) then we decide how to handle it, in your case you decided logging is enough. then you can log with the original exception that you caught.

However, What I'm saying is the root cause appears to be in the `unload` method, and the fix should to be applied there

RafyAmgadBenjamin: image

I have shown the exception to the user now 


regarding the `unload` method, it is actually not used anywhere in the code 
image

PR #596, opened 2020-08-24
use nginx for certificates in admin dashboard

OmarElawady:
### Description

This is to make the admin dashboard peertube solution compatible with the new image changes. It now uses nginx certbot to create a test certificates and serve the given domain name.

### DIFF
--- a/jumpscale/packages/tfgrid_solutions/chats/peertube.py
+++ b/jumpscale/packages/tfgrid_solutions/chats/peertube.py
@@ -155,27 +155,7 @@ class Peertube(GedisChatBot):
             raise StopChatFlow(f"Failed to create subdomain {self.domain} on gateway" f" {self.gateway.node_id} {_id}")
         self.threebot_url = f"https://{self.domain}"
 
-        # expose threebot container
-        _id = deployer.expose_address(
-            pool_id=self.pool_id,
-            gateway_id=self.gateway.node_id,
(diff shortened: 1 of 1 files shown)


jumpscale/packages/tfgrid_solutions/chats/peertube.py
```diff
@@ -155,27 +155,7 @@ def reservation(self):
...
-            solution_uuid=self.solution_id,
-            **self.solution_metadata,
-        )
-        success = deployer.wait_workload(_id, self)
-        if not success:
-            # solutions.cancel_solution(self.workload_ids)
-            raise StopChatFlow(f"Failed to create trc container on node {self.selected_node.node_id}" f" {_id}")
-
-        entrypoint = f'/usr/local/bin/startup.sh "{self.domain}" "{self.email}"'
+        entrypoint = f'/usr/local/bin/startup.sh "{self.domain}"'
```
AhmedSa-mir, line 158: Why did you remove the email parameter from the command?

AhmedSa-mir (approved):

jumpscale/packages/tfgrid_solutions/chats/peertube.py
OmarElawady: The only use for the email was to generate the certificate now nginx container creates it.

xmonader (approved):

Issue #636, opened 2020-08-25
add solution location to the soltution details for both admin panel and markerplace solutions 


hossnys:
need to add solution location at any these two windows of solution details , as many users are not familiar with how to get the farm name using explorer or explorer endpoints.  

![Screenshot from 2020-08-25 19-00-35](https://user-images.githubusercontent.com/26298009/91204728-c9380a80-e704-11ea-87b1-ba44b24d885e.png)

![Screenshot from 2020-08-25 19-00-51](https://user-images.githubusercontent.com/26298009/91204736-cb01ce00-e704-11ea-886a-075294ba1be2.png)

xmonader:
Do you mean hyperlink for node id to link to explorer node info?

waleedhammam:
> Do you mean hyperlink for node id to link to explorer node info?

I think the deployment location example: Belgium, Freefarm

hossnys:
I'm with Waleed opinion plus adding the farm name

OmarElawady:
Sometimes the node contains location and the farm doesn't like (node)[https://explorer.testnet.grid.tf/api/v1/nodes/3BwDVNRigau8Zk6BQkJt5PLxrHQnJ5NTfNvAtuubC1eS] and (farm)[https://explorer.testnet.grid.tf/api/v1/farms/71]. Is this a normal behaviour and is it okay to add only the farm info?
![Screenshot from 2020-08-27 11-38-26](https://user-images.githubusercontent.com/13040543/91654715-8632ba00-eaab-11ea-9a5f-676229085dfe.png)

Dina-Abd-Elrahman:
**Commit_ID** 1fcd8eec02bc4e39794e0ab22b722106e05ce7c4 , **Branch** development

![ff](https://user-images.githubusercontent.com/21218053/91865632-e2473b00-ec71-11ea-9452-d798b04d6806.png)

![ff1](https://user-images.githubusercontent.com/21218053/91865719-f854fb80-ec71-11ea-98f4-fe875f5e34e6.png)

PR #685, opened 2020-08-27
Development farm name in solutions

OmarElawady:
### Description

Add farm information to the solutions details.

## Related issues

#636

## Notes

1. I checked and in some cases the node contains location but the farm doesn't. This PR only adds the farm info (name and location).
2. It makes two api calls per node to fetch its farm which might slow down the listing in case there's a lot of solutions.

### DIFF
--- a/jumpscale/sals/reservation_chatflow/solutions.py
+++ b/jumpscale/sals/reservation_chatflow/solutions.py
@@ -10,11 +10,8 @@ K8S_SIZES = {
 class ChatflowSolutions:
     def get_node_farm(self, node_id):
         explorer = j.core.identity.me.explorer
-        try:
-            farm_id = explorer.nodes.get(node_id).farm_id
-            return str(explorer.farms.get(farm_id))
-        except:
-            return "Couldn't fetch farm"
+        farm_id = explorer.nodes.get(node_id).farm_id
+        return str(explorer.farms.get(farm_id))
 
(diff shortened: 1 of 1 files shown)


dmahmouali (approved):

jumpscale/sals/reservation_chatflow/solutions.py
```diff
@@ -8,6 +8,14 @@
 
 
 class ChatflowSolutions:
+    def get_node_farm(self, node_id):
+        explorer = j.core.identity.me.explorer
+        try:
+            farm_id = explorer.nodes.get(node_id).farm_id
+            return str(explorer.farms.get(farm_id))
+        except:
```
abom, line 16: Can we  be more specific here? what type of exceptions can occur?

jumpscale/sals/reservation_chatflow/solutions.py
```diff
@@ -575,6 +588,7 @@ def meta_filter(metadata):
                             "IPv4 Address": c_dict["ipv4"],
                             "IPv6 Address": c_dict["ipv6"],
                             "Node": c_dict["node"],
+                            "Farm": self.get_node_farm(c_dict["node"]),
```
abom, line 591: why we still use this "Titled" case as dict keys? If it's just for consistency with other keys, just ignore my comment :smile:

rkhamis (approved):

Issue #729, opened 2020-08-30
Solution expose when you don't have solutions it should show an error not an empty message


xmonader:
Solution expose when you don't have solutions it should show an error not an empty message with 

```
expose solutions:

EMPTY LIST
```

![image](https://user-images.githubusercontent.com/64129/91657766-d3bb2100-eac3-11ea-80d0-7628842a5943.png)

Dina-Abd-Elrahman:
**Comiit_ID** b6c2403cad6086756451c025dba8df700b82a644 , **Branch** development

![ff](https://user-images.githubusercontent.com/21218053/91663614-3de6bc80-eaea-11ea-9906-314eaf60de51.png)

PR #733, opened 2020-08-30
Added checking in solutions listing in solution expose

AhmedSa-mir:
### Description

Added checking in solutions listing in solution expose

### Related Issues

https://github.com/threefoldtech/js-sdk/issues/729

### DIFF
--- a/jumpscale/packages/tfgrid_solutions/chats/exposed.py
+++ b/jumpscale/packages/tfgrid_solutions/chats/exposed.py
@@ -46,7 +46,7 @@ class SolutionExpose(GedisChatBot):
 
     @chatflow_step(title="Solution to be exposed")
     def exposed_solution(self):
-        if len(self.sols.keys()) == 0:
+        if not self.sols:
             raise StopChatFlow(f"You don't have {self.kind} solutions")
         self.solution_name = self.single_choice(
             "Please choose the solution to expose", list(self.sols.keys()), required=True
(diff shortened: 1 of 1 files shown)


jumpscale/packages/tfgrid_solutions/chats/exposed.py
```diff
@@ -46,6 +46,8 @@ def solution_type(self):
 
     @chatflow_step(title="Solution to be exposed")
     def exposed_solution(self):
+        if len(self.sols.keys()) == 0:
```
m-motawea, line 49: maybe use `if not self.sols` ?

m-motawea (approved):

jumpscale/packages/tfgrid_solutions/chats/exposed.py
```diff
@@ -46,6 +46,8 @@ def solution_type(self):
 
     @chatflow_step(title="Solution to be exposed")
     def exposed_solution(self):
```
xmonader, line 48: can we check the length of all solutions in the beginning of the chatflow (and show only a list of what we have solutions for)

xmonader (changes requested):
I think we need to update how this works

1- we shouldn't start with the chatflow if we don't have solutions at all
2- we should filter out solution types that don't have any deployed instances.

xmonader (changes requested):
can we change the message of 
flist to be -> generic container
minio -> s3 storage

?

AhmedSa-mir:
> can we change the message of
> flist to be -> generic container
> minio -> s3 storage
> 
> ?

This should be a separate issue because then we need to change these fields in different places in code

PR #806, opened 2020-09-01
fix url for mainnet farm management

DylanVerstraete:


jumpscale/packages/farmmanagement/frontend/store/farmmanagement.js
```diff
@@ -22,12 +22,16 @@ export default {
   },
   actions: {
     getTfgridUrl: async context => {
-      var response = await tfService.getExplorer();
-      var url = JSON.parse(response.data).data.url
-      if (!url.startsWith('http')) {
-        url = `https://${url}/api/v1`;
```
m-motawea, line 28: you can just use this endpoint to get the explorer url `/admin/actors/identity/get_explorer_url` on development branch
result: 
```
"{\"url\": \"https://explorer.testnet.grid.tf/api/v1\"}"
```

DylanVerstraete: @m-motawea this is already happening right now: 
https://github.com/threefoldtech/js-sdk/blob/42e1e1c93b0b3b9992c4b4cb746689abb2cb6b64/jumpscale/packages/admin/actors/admin.py#L34
https://github.com/threefoldtech/js-sdk/blob/42e1e1c93b0b3b9992c4b4cb746689abb2cb6b64/jumpscale/packages/farmmanagement/frontend/services/farmmanagementServices.js#L5

DylanVerstraete: This just returns `explorer.tesnet.grid.tf` or `explorer.grid.tf` it does not return what comes after that eg: `/api/v1` in the case of testnet and `/explorer` in the case of mainnet....

m-motawea: sorry seems I commented on a wrong line. you don't need to hardcode "/explorer" or "/api/v1". the above actor returns the correct one as you can see in the response

DylanVerstraete: @m-motawea no it does not. it return : `"{\"data\": {\"type\": \"testnet\", \"url\": \"explorer.testnet.grid.tf\"}}"`

m-motawea: image

this is the latest on development.

PR #985, opened 2020-09-03
branch for grammar corrections all

sasha-astiadi:
some grammar

### DIFF
--- a/docs/api/jumpscale/packages/marketplace/chats/exposed.html
+++ b/docs/api/jumpscale/packages/marketplace/chats/exposed.html
@@ -61,7 +61,7 @@ class SolutionExpose(BaseSolutionExpose, MarketPlaceChatflow):
 
         gateways = deployer.list_all_gateways(self.solution_metadata[&#34;owner&#34;])
         if not gateways:
-            raise StopChatFlow(&#34;There are no available gateways in the farms bound to your pools&#34;)
+            raise StopChatFlow(&#34;There is no available gateway in the farms bound to your capacity pools&#34;)
 
         # add managed domains
         gateway_id_dict = {}
@@ -170,7 +170,7 @@ any extra kwargs that is passed while creating the session
 
         gateways = deployer.list_all_gateways(self.solution_metadata[&#34;owner&#34;])
         if not gateways:
-            raise StopChatFlow(&#34;There are no available gateways in the farms bound to your pools&#34;)
+            raise StopChatFlow(&#34;There is no available gateway in the farms bound to your capacity pools&#34;)
 
         # add managed domains
         gateway_id_dict = {}
@@ -302,7 +302,7 @@ any extra kwargs that is passed while creating the session
 
         gateways = deployer.list_all_gateways(self.solution_metadata[&#34;owner&#34;])
         if not gateways:
-            raise StopChatFlow(&#34;There are no available gateways in the farms bound to your pools&#34;)
+            raise StopChatFlow(&#34;There is no available gateway in the farms bound to your capacity pools&#34;)
 
         # add managed domains
         gateway_id_dict = {}
--- a/jumpscale/packages/marketplace/chats/blog.py
+++ b/jumpscale/packages/marketplace/chats/blog.py
@@ -7,17 +7,17 @@ from jumpscale.sals.chatflows.chatflows import chatflow_step
 
 class BlogDeploy(Publisher):
 
-    title = "Deploy Blog"
-    SOLUTION_TYPE = "blog"  # chatflow used to deploy the solution
+    title = "Deploy a Blog"
+    SOLUTION_TYPE = "Blog"  # chatflow used to deploy the solution
 
-    @chatflow_step(title="blog Setup")
+    @chatflow_step(title="Blog Set Up")
     def configuration(self):
         form = self.new_form()
         title = form.string_ask("Title", required=True)
-        url = form.string_ask("Repository url", required=True)
+        url = form.string_ask("Repository URL", required=True)
         branch = form.string_ask("Branch", required=True)
-        form.ask("Set configuration")
-        self.user_email = self.user_info()["email"]
+        form.ask("Set Configuration")
+        self.user_email = self.user_info()["E-mail"]
         self.envars = {
             "TYPE": "blog",
             "NAME": "entrypoint",
--- a/jumpscale/packages/marketplace/chats/exposed.py
+++ b/jumpscale/packages/marketplace/chats/exposed.py
@@ -21,7 +21,7 @@ class SolutionExpose(BaseSolutionExpose, MarketPlaceChatflow):
         self.username = self.user_info()["username"]
         self.solution_metadata["owner"] = self.username
 
-    @chatflow_step(title="Solution type")
+    @chatflow_step(title="Solution Type")
     def solution_type(self):
         self._deployment_start()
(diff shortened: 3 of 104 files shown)


docs/api/jumpscale/packages/marketplace/chats/exposed.html
```diff
@@ -61,7 +61,7 @@ <h1 class="title">Module <code>jumpscale.packages.marketplace.chats.exposed</cod
 
         gateways = deployer.list_all_gateways(self.solution_metadata[&#34;owner&#34;])
         if not gateways:
-            raise StopChatFlow(&#34;There are no available gateways in the farms bound to your pools&#34;)
+            raise StopChatFlow(&#34;There is no available gateway in the farms bound to your capacity pools&#34;)
```
waleedhammam, line 64: This is a generated file we shouldn't change here

docs/api/jumpscale/packages/marketplace/chats/exposed.html
```diff
@@ -170,7 +170,7 @@ <h2 class="section-title" id="header-classes">Classes</h2>
...
```
waleedhammam, line 173: also here, same as all html files

jumpscale/packages/marketplace/chats/blog.py
```diff
@@ -7,17 +7,17 @@
 
 class BlogDeploy(Publisher):
 
-    title = "Deploy Blog"
-    SOLUTION_TYPE = "blog"  # chatflow used to deploy the solution
+    title = "Deploy a Blog"
+    SOLUTION_TYPE = "Blog"  # chatflow used to deploy the solution
```
waleedhammam, line 11: SOLUTION_TYPE is an enum, contains the types of the chatflows all lowered case shouldn't be changed(python semantics)

jumpscale/packages/marketplace/chats/threebot_advanced.py
```diff
@@ -23,7 +23,7 @@ def set_threebot_name(self):
             for sol in threebot_solutions:
                 if sol["Name"] == self.solution_name:
                     valid = False
-                    self.md_show("The specified solution name already exists. please choose another.")
+                    self.md_show("The specified solution name already exists. please choose another name name.")
                     break
```
waleedhammam, line 27: Why is name name that? Typo maybe?

jumpscale/packages/marketplace/chats/website.py
```diff
@@ -10,14 +10,14 @@ class WebsiteDeploy(Publisher):
...
     def configuration(self):
         form = self.new_form()
         title = form.string_ask("Title", required=True)
-        url = form.string_ask("Repository url", required=True)
+        url = form.string_ask("Repository URL", required=True)
         branch = form.string_ask("Branch", required=True)
-        form.ask("Set configuration")
-        self.user_email = self.user_info()["email"]
+        form.ask("Set Configuration")
+        self.user_email = self.user_info()["E-mail"]
```
waleedhammam, line 20: this is user_info() method that has information for username and email, changing that will cause index error, why is that?

jumpscale/packages/marketplace/chats/wiki.py
```diff
@@ -19,7 +19,7 @@ def configuration(self):
         form.ask("Set configuration")
         self.user_email = self.user_info()["email"]
         self.envars = {
-            "TYPE": "wiki",
+            "TYPE": "Wiki",
```
waleedhammam, line 22: this env variable will be delegated to the flist, changing that will cause flist failure!

jumpscale/packages/tfgrid_solutions/chats/publisher.py
```diff
@@ -80,11 +80,11 @@ def publisher_info(self):
     @chatflow_step(title="Solution Settings")
     def configuration(self):
         form = self.new_form()
-        ttype = form.single_choice("Choose the type", options=["wiki", "www", "blog"], default="wiki", required=True)
+        ttype = form.single_choice("Choose Your Publicattion Type", options=["Wiki", "www", "Blog"], default="Wiki", required=True)
```
waleedhammam, line 83: this result will be affecting a lot of checks, we shouldn't change that

waleedhammam (changes requested):
Thanks for the grammar review. but this PR has breaking changes in variables names, I added some of them the rest follows too

xmonader:
Wow. Lot of grammar mistakes. Thank you! This PR made me think of having an easier way for collaboration and guard the codebase, because we need to distinguish between strings used for human interaction and strings used in data exchange between computers, a slight change in the wrong place in the strings used between computer services, breaks functionality 

https://github.com/threefoldtech/js-sdk/issues/986

Here's an example of data we store on the grid
image

- as you can see things like solutions, their case means something.
- data that is passed when launching containers and checks there expects the input to be of certain format and names
- things like `self.user_info()["E-mail"]` breaks our code, that data that's coming from 3bot login has the email in lowercase.
- I'll make sure this PR to be used as a reference for the things it can be used, and for the technical definitions used (these will be reviewed)

jumpscale/packages/marketplace/chats/exposed.py
```diff
@@ -68,27 +68,27 @@ def domain(self):
             retry = False
             while True:
                 domain = self.string_ask(
-                    f"Please specify the sub domain name you wish to bind to. will be (subdomain).{self.domain}",
+                    f"Please specify the sub domain name you wish to bind to. This will be a (subdomain).{self.domain}",
```
ranatrk, line 71: the example is shown here , so probably should be `This will be (subdomain).{self.domain}` without "a"

sasha-astiadi:
Thank you! Yes I was aware of that some are not even texts but strings... But when I replaced some stuff, for example, email to e-mail, I was doing it with search & find, and it automatically replaced strings that I didn't want to replace... Sorry!

I think I caught most of unwanted replacement, for example, timedout to time out.. I saw that when I did it in batch, it tried to change some timedout string hihi.

Hope this is useful. I'm not sure how you guys can put this input without breaking the codes. 
Are capslock ; case sensitivity matter in python?

jumpscale/packages/marketplace/chats/threebot_advanced.py
sasha-astiadi: Yes typo

jumpscale/packages/marketplace/chats/website.py
sasha-astiadi: also it was automated change from search and replace feature, sorry

jumpscale/packages/marketplace/chats/wiki.py
sasha-astiadi: also it was automated change from search and replace feature, sorry

jumpscale/packages/tfgrid_solutions/chats/publisher.py
sasha-astiadi: if its a multiple choice, a beginning of a sentence, please always ALWAYS big first letter

Blog
Wiki
Website (dont call it www)

jumpscale/packages/marketplace/chats/exposed.py
sasha-astiadi: a beginning of the sentence after a "." should ALWAYS have big first letter. 

example:

I do this sentence. Now the first letter in 'now' on the second sentence is always big.

xmonader:
PR processed in #994

Issue #1071, opened 2020-09-08
prepare for mainnet upgrade with explorer API endpoint


xmonader:
- the explorer endpoint for mainnet is going to be https://explorer.grid.tf/api/v1

- how do we upgrade for existing identities on the configurations?
- should we define the urls as js config and reuse them from there? that will allow automigrate in the future?

waleedhammam:
Blocked on the upgrade of mainnet url

PR #1096, opened 2020-09-09
add upgrade mainnet identity

waleedhammam:
### Changes

- auto upgrade mainnet
- save explorers in js-configs to reuse later
- After migration is complete we will trigger this update by setting new config `has_migrated=True`

### Related Issues

- https://github.com/threefoldtech/js-sdk/issues/1071

### DIFF
--- a/jumpscale/core/identity/__init__.py
+++ b/jumpscale/core/identity/__init__.py
@@ -101,9 +101,10 @@ class Identity(Base):
 
             # Backward compitablity (Update mainnet url)
             # Create new config has_migrated = True after mainnet update
-            if js_config.get("has_migrated_explorer_url", False):
+            if not js_config.get("has_migrated_explorer_url", False):
                 if urlparse(self.explorer_url).hostname == urlparse(EXPLORER_URLS["mainnet"]).hostname:
                     self.explorer_url = EXPLORER_URLS["mainnet"]
+                    js_config.set("has_migrated_explorer_url", True)
 
             if self.explorer_url:
                 self._explorer = ex_factory.get_by_url_and_identity(self.explorer_url, identity_name=self.instance_name)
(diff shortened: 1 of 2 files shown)


jumpscale/core/identity/__init__.py
```diff
@@ -87,6 +98,13 @@ def explorer(self):
             from jumpscale.clients.explorer import export_module_as  # Import here to avoid circular imports
 
             ex_factory = export_module_as()
+
+            # Backward compitablity (Update mainnet url)
+            # Create new config has_migrated = True after mainnet update
+            if js_config.get("has_migrated", False):
```
abom, line 104: Can we rename it to a more specific name, to be related to this explorer URL change.
Also, we can add a note about better network selection implementation.

jumpscale/core/identity/__init__.py
```diff
@@ -87,6 +98,13 @@ def explorer(self):
             from jumpscale.clients.explorer import export_module_as  # Import here to avoid circular imports
 
             ex_factory = export_module_as()
+
+            # Backward compitablity (Update mainnet url)
+            # Create new config has_migrated = True after mainnet update
+            if js_config.get("has_migrated_explorer_url", False):
+                if urlparse(self.explorer_url).hostname == urlparse(EXPLORER_URLS["mainnet"]).hostname:
+                    self.explorer_url = EXPLORER_URLS["mainnet"]
```
abom, line 106: We should set `has_migrated_explorer_url` to `True` here, so we don't do it again.

jumpscale/core/identity/__init__.py
```diff
@@ -87,6 +98,13 @@ def explorer(self):
             from jumpscale.clients.explorer import export_module_as  # Import here to avoid circular imports
 
             ex_factory = export_module_as()
+
+            # Backward compitablity (Update mainnet url)
+            # Create new config has_migrated = True after mainnet update
+            if js_config.get("has_migrated_explorer_url", False):
```
abom, line 104: This should be `if not...`

waleedhammam:
TO MERGE AFTER UPGRADING MAINNET URL

xmonader (approved):

Issue #1171, opened 2020-09-15
abnormal behavior when peertube deploy fail, redirect to list peertube page 


dmahmouali:
- when deploying peertube on https://demo2.testnet.grid.tf/ and https://demo.testnet.grid.tf/, deployment fails with error alert then redirects directly to the listing page without any user action.

OmarElawady:
The chatflow listens for payloads from the backend and depending on their type,  it disables or enables the next button. The issue is that calling `self.user_info` is a method that sends a payload to the chatflow with a type for which the next button is enabled. If the user clicks on the next again, it registers two transitions and that causes the step to come to be skipped. I moved it to the beginning of the step so that the next button don't get re-enabled.

Pishoy:
verified
![image](https://user-images.githubusercontent.com/21987477/93341010-e48ec500-f82d-11ea-9d31-fdea508837ed.png)

PR #1178, opened 2020-09-15
fetch user info at the beginning

OmarElawady:
### Description

The chatflow listens for payloads from the backend and depending on their type, it disables or enables the next button. The issue is that calling self.user_info is a method that sends a payload to the chatflow with a type for which the next button is enabled. If the user clicks on the next again, it registers two transitions and that causes the step to come to be skipped. I moved it to the beginning of the step so that the next button don't get re-enabled.

### Related Issues

#1171


waleedhammam (approved):
LGTM

PR #1201, opened 2020-09-17
Fix farm node deletion

DylanVerstraete:


jumpscale/packages/farmmanagement/frontend/components/nodestable/nodestable.js
```diff
@@ -185,11 +190,9 @@ module.exports = new Promise(async (resolve, reject) => {
                                 type: "error",
                             }
                         }
-                        setTimeout(() => {
-                            this.deleteNodeFarmAlert = undefined
-                            this.openDeleteModal = false
-                            this.getNodes(node.farm.id)
-                        }, 2000)
+                        this.deleteNodeFarmAlert = undefined
```
zaibon, line 193: I think removing the `setTimeout` there mean the user will not have the time to see the error/success message.
We should move the v-alert outside of the dialog window then

zaibon (changes requested):

zaibon (approved):

Issue #1246, opened 2020-09-22
adding a courrpted package  (non exists)  will courrpt the  threebot server


Dina-Abd-Elrahman:
- in threebot server when I added non exist package EX: `blabla` it prints an error for you that there is package.toml file for this package and cannot add it, But if you try to list a packages with them using this command `j.servers.threebot.default.packages.list_all()` you will find all wrong packages you add.

![ff1](https://user-images.githubusercontent.com/21218053/93869465-1db2b380-fccc-11ea-80f6-ff49c035d087.png)

and if you stop the threebot server and try to start it again you got this error

![ff](https://user-images.githubusercontent.com/21218053/93869301-e93ef780-fccb-11ea-85bb-a25d217e4633.png)

**Commit_ID** a3239ed343ad62283f049ce65f2ca564684cc8f4 , **Branch** development

RafyAmgadBenjamin:
### fixed on branch 
`development_add_package_issue`
### commit id 
90c100d9d26cb3f18ee2993453bc4ff0131248a9

Dina-Abd-Elrahman:
**Commit_ID** 69ef97d8e49fb62fb6b7b41bbcbc86cc32faff0f , **Branch** development

![ff](https://user-images.githubusercontent.com/21218053/93988077-9a0acc80-fd88-11ea-9e5f-1d7a9c00961c.png)

PR #1248, opened 2020-09-22
fix an issue related to adding wrong package

RafyAmgadBenjamin:
### Description

Fixing an issue related to adding  a not correct package
### Changes
I have update the code to make the code to do installing the package and if it succeeded i will add the package to our packages if not I will raise exception

List of changes this PR includes
- jumpscale/servers/threebot/threebot.py

### Related Issues
https://github.com/threefoldtech/js-sdk/issues/1246

### DIFF
--- a/jumpscale/servers/threebot/threebot.py
+++ b/jumpscale/servers/threebot/threebot.py
@@ -373,14 +373,6 @@ class PackageManager(Base):
         if not any([path, giturl]) or all([path, giturl]):
             raise j.exceptions.Value("either path or giturl is required")
 
-        for package_name in self.packages:
-            package = self.get(package_name)
-            ## TODO: why do we care if the path is the same and giturl is the same? adding it 100 times should just add it once?
-            # if path and path == package.path:
-            #     raise j.exceptions.Value("Package with the same path already exists")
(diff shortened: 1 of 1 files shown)


AhmedSa-mir (approved):

jumpscale/servers/threebot/threebot.py
```diff
@@ -407,15 +399,18 @@ def add(self, path: str = None, giturl: str = None, **kwargs):
...
-        if self.threebot.started:
-            self.install(package)
-            self.threebot.nginx.reload()
+        try:
+            if self.threebot.started:
+                self.install(package)
+                self.threebot.nginx.reload()
+            self.packages[package.name] = {"name": package.name, "path": package.path, "giturl": package.giturl}
+        except Exception as e:
+            j.exceptions.Runtime(f"Error happend during adding {package.name}, {str(e)}")
```
abom, line 412: I think this will mask any thing that can go wrong, and will only give us a `Runtime` exception.

Can we change the order of adding the package to `self.packages` without try/except?

RafyAmgadBenjamin: done

abom (approved):

PR #1272, opened 2020-09-25
add set_default on identity object

xmonader:
instead of creating identity and then  go to `j.core.identity.set_default("newme")` we can `identityobj.set_default()`

### DIFF
--- a/jumpscale/core/identity/__init__.py
+++ b/jumpscale/core/identity/__init__.py
@@ -171,6 +171,11 @@ class Identity(Base):
         self.save()
         return tid
 
+    def set_default(self):
+        from jumpscale.loader import j
+
+        return j.core.identity.set_default(self.instance_name)
+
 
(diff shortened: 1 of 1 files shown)


jumpscale/core/identity/__init__.py
```diff
@@ -171,6 +171,11 @@ def register(self, host=None):
         self.save()
         return tid
 
+    def set_default(self):
+        from jumpscale.loader import j
+
+        return j.core.identity.set_default(self.instance_name)
```
dmahmouali, line 177: Is that cahnge need `j.core.identity.me.save()` or not ?

dmahmouali: no need to use `save`.

dmahmouali (approved):

waleedhammam (approved):

PR #1289, opened 2020-09-28
add nginx test coverage

BolaNasr:


tests/tools/nginx/test_nginx.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        nginx_instance = self._get_instance()
+        nginx_instance.save()
+        nginx_instance.start()
+        j.logger.info("NGINX server started")
+
+        self.assertTrue(nginx_instance.is_running())
+
+        nginx_instance.stop()
+        time.sleep(2)
```
AhmedHanafy725, line 32: Instead of sleep, can we wait for port 80 to be released with timeout 2 sec?

AhmedHanafy725 (commented):
both tests are failing in github actions, can u check them?

AhmedHanafy725 (approved):

PR #1297, opened 2020-09-28
threebot server

Dina-Abd-Elrahman:
https://github.com/threefoldtech/js-sdk/issues/1267
- Add test for threebot server.
- Fix identity tests for running locally with existing identity.


tests/servers/threebot/test_package_manager.py
```diff
@@ -0,0 +1,103 @@
+import os
+from gevent import monkey
+from random import randint
+from jumpscale.loader import j
+from tests.base_tests import BaseTests
+
+monkey.patch_all(subprocess=False)
```
abom, line 7: I think monkey-patching should be done in tests module level, otherwise it would be defined many times (also, we need monkey patching to be enabled for all tests, not only threebot).

AhmedHanafy725: I think the right place for this is [conftest.py](https://github.com/threefoldtech/js-sdk/blob/development/conftest.py) as it called while discovering the tests.

tests/servers/threebot/test_threebot.py
```diff
@@ -0,0 +1,124 @@
...
+        self.info(" * Check that nginx server connection works successfully and right port.")
+        self.assertTrue(j.sals.nettools.tcp_connection_test("localhost", 80, 2))
+        self.assertTrue(j.sals.nettools.tcp_connection_test("localhost", 443, 2))
+
+        self.info("*** redis server ***")
+        _, redis_output, error = j.sals.process.execute(" ps -aux | grep -v grep | grep redis")
+        self.assertTrue(redis_output)
+        self.info(" * Check that redis server connection  works successfully and right port.")
+        self.assertTrue(j.sals.nettools.tcp_connection_test("localhost", 6379, 2))
+
```
AhmedHanafy725, line 47: You can also check gedis_http server running on 8000.

AhmedHanafy725 (commented):
It is failing on github actions, can you check please?

waleedhammam (approved):

PR #1315, opened 2020-09-29
Add identity cover test

BolaNasr:


AhmedHanafy725 (commented):
Can you add test for registering fake identity?

AhmedHanafy725 (commented):
These tests need `teardown` to revert what is happening in the tests.

tests/core/identity/test_identity.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        identity = self._get_instance()
+        identity.register()
+        j.logger.info("Identity registered")
+
+        self.assertTrue(os.getenv("tname"), j.core.identity.find(self.instance_name).tname)
+
+        j.core.identity.delete(self.instance_name)
+
+        self.assertTrue(type(None), j.core.identity.find(self.instance_name))
```
AhmedHanafy725, line 76: we can use `assertIsNone` instead/

tests/core/identity/test_identity.py
```diff
@@ -0,0 +1,102 @@
...
+    def test004_register_fake_identity(self):
+        """Test case for delete my identity.
+        **Test scenario**
+        #. Get identity instance.
+        #. Register an instance.
+        #. Delete my identity.
+        #. Search if my identity exist or not.
+        """
+        self.instance_name = j.data.random_names.random_name()
+        self.me = j.core.identity.me
```
AhmedHanafy725, line 87: Why we don't have this line in the `setUp` method?

tests/core/identity/test_identity.py
```diff
@@ -17,7 +17,7 @@ def _get_instance(self):
                 explorer_url="https://explorer.testnet.grid.tf/api/v1",
             )
         else:
-            raise Exception("Please add (name, Email, words) of your 3bot identity as Env ")
+            raise Exception("Please add (tname, email, words) of your 3bot identity as Env ")
```
xmonader, line 22: as Env means as in .env file or as environment variables? and if as environment variables, they should be all upper case

AhmedHanafy725 (approved):

PR #1324, opened 2020-09-29
parameterize zos in identity

xmonader:
parameterize zos  in identity so if we want to switch identity at certain point we can


jumpscale/sals/zos/__init__.py
```diff
@@ -1,4 +1,34 @@
...
+    Returns:
+        [type]: [description]
+    """
+    global INSTANCES, ME_INSTANCE
+    from jumpscale.loader import j
     from .zos import Zosv2
 
-    return Zosv2()
+    if not identity_name:
+        identity_name = j.core.identity.me.instance_name
```
waleedhammam, line 20: I think If I have no default identity this one gonna crash ?

xmonader: Addressed, also handled the case of using identity_name that doesn't exist

waleedhammam (approved):
When merging, we should make sure to merge back with development and make the new usage,  some new changes on were introduced after the pr was created

xmonader:
I updated to the latest please let's give it some prio

abom (approved):

PR #1352, opened 2020-10-04
Add Ubuntu server script to test gateway failover

OmarElawady:
### Description

Add Ubuntu server script to test gateway failover.

### DIFF
--- a/examplescripts/ubuntu_server.py
+++ b/examplescripts/ubuntu_server.py
@@ -1,0 +1,163 @@
+from jumpscale.loader import j
+from time import sleep
+import uuid
+
+zos = j.sals.zos
+
+payment_detail = zos.pools.create(cu=100, su=100, farm="freefarm", currencies=["TFT"])
+print(payment_detail)
+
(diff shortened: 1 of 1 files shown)


examplescripts/ubuntu_server.py
```diff
@@ -0,0 +1,181 @@
...
+    print(payment_detail)
+
+    wallet = j.clients.stellar.get(wallet_name)
+    txs = zos.billing.payout_farmers(wallet, payment_detail)
+    pool = zos.pools.get(payment_detail.reservation_id)
+    while pool.cus == 0:
+        pool = zos.pools.get(payment_detail.reservation_id)
+        sleep(1)
+
+    print("compute units available:", pool.cus)
```
abom, line 47: for consistency, could we always use logger?

OmarElawady: Those prints were only for debugging so I removed them. I also changed the remaining print to use logger, except for the network configuration part to make copy pasting easier.

abom (approved):

PR #1355, opened 2020-10-04
update taiga client with more methods to get and export data

RafyAmgadBenjamin:
### Description

Adding more methods for Taiga to get and export data into markdown files
### Changes

`jumpscale/clients/taiga/taiga.py`

### Related Issues

https://github.com/threefoldtech/home/issues/773

### DIFF
--- a/jumpscale/clients/taiga/taiga.py
+++ b/jumpscale/clients/taiga/taiga.py
@@ -260,7 +260,7 @@ class TaigaClient(Client):
         """
         return self.api.user_stories.list(assigned_to=user_id)
 
-    def render_issues_with_details(self, text="Issues", issues=None):
+    def __render_issues_with_details(self, text="Issues", issues=None):
         """Get issues subject and append them to text
         
         Args:
@@ -271,15 +271,33 @@ class TaigaClient(Client):
             str: string contains the issues subject.
         """
         for issue in issues:
-            text += f"- **Subject:** {issue.subject} \n"
-            text += f"  - **Created Date:** {issue.created_date} \n"
-            text += f"  - **Due Date:** {issue.due_date} \n"
-            text += f"  - **Owner Name:** {issue.owner_extra_info.get('full_name_display','unknown')} \n"
-            text += f"  - **Owner Email:** {issue.owner_extra_info.get('email','unknown')} \n"
-            text += f"  - **Project:** {issue.project_extra_info.get('name','unknown')} \n"
+            text += f"- **Subject:** {issue.get('subject')} \n"
+            text += f"  - **Created Date:** {issue.get('created_date')} \n"
+            text += f"  - **Due Date:** {issue.get('due_date')} \n"
+            text += f"  - **Owner Name:** {issue.get('owner_name','unknown')} \n"
+            text += f"  - **Owner Email:** {issue.get('owner_mail','unknown')} \n"
+            text += f"  - **Project:** {issue.get('project','unknown')} \n"
         return text
 
-    def export_all_issues_details(self, path="/tmp/issues_titles.md"):
+    def __get_issues_required_data(self, issues):
+        """Get the required data from issues to be used later in export(render)
+
+        Returns:
+            List: List of issues required details to be used in export(render).
+        """
+        ALL_ISSUES_TEMPLATES = list()
+        for issue in issues:
+            SINGLE_PROJECT_TEMPLATE = dict()
+            SINGLE_PROJECT_TEMPLATE["subject"] = issue.subject
+            SINGLE_PROJECT_TEMPLATE["created_date"] = issue.created_date
+            SINGLE_PROJECT_TEMPLATE["due_date"] = issue.due_date
+            SINGLE_PROJECT_TEMPLATE["owner_name"] = issue.owner_extra_info.get("full_name_display", "unknown")
+            SINGLE_PROJECT_TEMPLATE["owner_mail"] = issue.owner_extra_info.get("email", "unknown")
+            SINGLE_PROJECT_TEMPLATE["project"] = issue.project_extra_info.get("name", "unknown")
+            ALL_ISSUES_TEMPLATES.append(SINGLE_PROJECT_TEMPLATE)
+        return ALL_ISSUES_TEMPLATES
+
+    def export_all_issues_details(self, path="/tmp/issues_details.md"):
         """Export all the issues subjects in a markdown file
         
         Args:
@@ -290,7 +308,9 @@ class TaigaClient(Client):
 ### Issues \n
 """
         issues = self.list_all_issues()
-        text = self.render_issues_with_details(text=text, issues=issues)
+        ALL_ISSUES_TEMPLATES = self.__get_issues_required_data(issues)
+
+        text = self.__render_issues_with_details(text=text, issues=ALL_ISSUES_TEMPLATES)
 
         j.sals.fs.write_file(path=path, data=text)
 
@@ -305,10 +325,15 @@ class TaigaClient(Client):
         text = """
 ### Issues  Per Project \n
 """
+        ALL_PROJECTS_TEMPLATES = list()
         for project in projects:
-            text += f"##### {project.name} \n"
-            # text += self.render_issues_with_details(title=text, issues=project.list_issues())
-            for issue in project.list_issues():
+            SINGLE_PROJECT_TEMPLATE = dict()
+            SINGLE_PROJECT_TEMPLATE["name"] = project.name
+            SINGLE_PROJECT_TEMPLATE["issues"] = project.list_issues()
+            ALL_PROJECTS_TEMPLATES.append(SINGLE_PROJECT_TEMPLATE)
+        for project in ALL_PROJECTS_TEMPLATES:
+            text += f"##### {project.get('name')} \n"
+            for issue in project.get("issues"):
                 text += f"- {issue.subject} \n"
 
         j.sals.fs.write_file(path=path, data=text)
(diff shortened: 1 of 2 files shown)


jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,137 @@ def _get_project_status(project_statuses, status):
...
+        Args:
+            path (str): The path of exported markdown file.
+            
+        """
+        users_stories = self.list_all_user_stories()
+        text = f"""
+### All User Stories \n
+"""
+        for story in users_stories:
+            text += f"#### {story.subject} \n"
```
xmonader, line 344: Can we make a SINGLE_STORY_TEMPLATE variable outside and render it with each story here?

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,137 @@ def _get_project_status(project_statuses, status):
...
+        
+        Args:
+            path (str): The path of exported markdown file.
+            
+        """
+        projects = self.list_all_projects()
+        text = """
+### Issues  Per Project \n
+"""
+        for project in projects:
```
xmonader, line 303: can we make SINGLE_PROJECT_TEMPLATE variable and use it here instead of writing the data on the go?

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,137 @@ def _get_project_status(project_statuses, status):
...
+        
+        Args:
+            user_id (int): id of the user.
+            
+        Returns:
+            List: List of user stories.
+        """
+        return self.api.user_stories.list(assigned_to=user_id)
+
+    def get_issues_export_text(self, prepend_text="Issues", issues=None):
```
xmonader, line 263: The name here is confusing should be something like `render_issues_with_title(self, title, issues)`

jumpscale/clients/taiga/taiga.py
RafyAmgadBenjamin: I am not writing the data on the fly, I do that in the last line of the method 
image

jumpscale/clients/taiga/taiga.py
RafyAmgadBenjamin: you want me to change the name of the variable, or make a global variable and be used with every export method

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+
+    def list_all_issues(self, user_id=""):
+        """
+        List all issues for specific user if you didn't pass user_id will list all the issues
+        
+        Args:
+            user_id (int): id of the user.
+
+        Returns:
+            List: List of issues.
```
abom, line 229: Can we specify if issue is an object of which type to be more clear?

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+        """
+        return self.api.issues.list(assigned_to=user_id)
+
+    def list_all_projects(self):
+        """
+        List all projects
+        
+        Returns:
+            List: List of projects.
```
abom, line 238: Here too, it applies for all return types and arguments docstring, it's better to specify the type as a list of what?

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+        """
+        return self.api.projects.list()
+
+    def list_all_milestones(self):
+        """
+        List all milestones
+        
+        Returns:
+            List: List of milestones.
```
abom, line 247: the same as https://github.com/threefoldtech/js-sdk/pull/1355/files#r499798935.

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+
+    def list_all_user_stories(self, user_id=""):
+        """
+        List all user stories for specific user if you didn't pass user_id will list all the available user stories
+        
+        Args:
+            user_id (int): id of the user.
+            
+        Returns:
+            List: List of user stories.
```
abom, line 259: the same as https://github.com/threefoldtech/js-sdk/pull/1355/files#r499798935 too.

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+
+    def __render_issues_with_details(self, text="Issues", issues=None):
+        """Get issues subject and append them to text
+        
+        Args:
+            text (str): the text that the method will append to it.
+            issues (List): list of all issues that we will get the subject from .
+            
+        Returns:
+            str: string contains the issues subject.
```
abom, line 271: It's better to export to one of the known formats and to specify this in docstring too, what's the current format of this export? markdown?

IMHO, csv like formats would be a good choice, as this export would be both human and machine readable.

abom: Also, it's better to use a template file here if needed (I think won't be needed in case of csv for example).

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+            path (str): The path of exported markdown file.
+            
+        """
+        text = f"""
+### All User Stories \n
+"""
+        All_STORIES_TEMPLATE = list()
+        All_STORIES_TEMPLATE = self.__get_stories_required_data()
+
+        for story in All_STORIES_TEMPLATE:
```
abom, line 397: I think if we are going to use markdown, it's better to use a template (jinja template for example) instead of doing it like this.

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+        """Export all the user stories in a markdown file
+        
+        Args:
+            path (str): The path of exported markdown file.
+            
+        """
+        text = f"""
+### All User Stories \n
+"""
+        All_STORIES_TEMPLATE = list()
```
abom, line 394: what's the need for this line? already overwritten the line after.

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+        All_STORIES_TEMPLATE = self.__get_stories_required_data()
```
abom, line 395: why do we use all UPPER_CASE letter for a local variable?

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +217,190 @@ def _get_project_status(project_statuses, status):
...
+        text = self.__render_issues_with_details(text=text, issues=ALL_ISSUES_TEMPLATES)
+        j.sals.fs.write_file(path=path, data=text)
+
+    def __get_stories_required_data(self):
+        """Get the required data from user stories to be used later in export(render)
+
+        Returns:
+            List: List of stories required details to be used in export(render).
+        """
+        All_STORIES_TEMPLATE = list()
```
abom, line 369: why do we use all UPPER_CASE letter for a local variable?

abom:
Some comments that's not related to current changes but related to taiga client:

1- `super.__init__` need to take args & kwargs:

https://github.com/threefoldtech/js-sdk/blob/c006028b0668185b848b5fcd238c8639bd4e951d/jumpscale/clients/taiga/taiga.py#L16-L18


2- No need to implement own `__hash__` if not needed (just double check, I don't know for sure).

https://github.com/threefoldtech/js-sdk/blob/c006028b0668185b848b5fcd238c8639bd4e951d/jumpscale/clients/taiga/taiga.py#L20-L21

3- `api` property wouldn't get re-computed in case the token or any other info was changed, this can be solved by many ways mentioned [here in the docs](https://threefoldtech.github.io/js-ng/wiki/#/./baseclasses?id=data-updates-and-computed-fields):

https://github.com/threefoldtech/js-sdk/blob/c006028b0668185b848b5fcd238c8639bd4e951d/jumpscale/clients/taiga/taiga.py#L24-L35

jumpscale/clients/taiga/taiga.py
RafyAmgadBenjamin: you are right,  it is much better to use jinja template but since the markdown is simple we are not in a bad need of it, if the markdown got more complex we have to use jinja template.

jumpscale/clients/taiga/taiga.py
RafyAmgadBenjamin: The task was to export the data to markdown files.

RafyAmgadBenjamin:
regarding this (No need to implement own __hash__ if not needed (just double check, I don't know for sure).)

I don't know the case either @xmonader, do you have any insights here ?

jumpscale/clients/taiga/taiga.py
RafyAmgadBenjamin: we are just listing here

abom:
Review still in progress...

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +224,302 @@ def _get_project_status(project_statuses, status):
...
+        Args:
+            path (str): The path of exported markdown file.
+            
+        """
+        self.text = """
+### Issues  Per Project \n
+"""
+
+        greenlet = gevent.spawn(self.map_render_issues_per_project, with_details)
+        gevent.joinall([greenlet])
```
abom, line 415: Actually, this is just like doing it outside a greenlet, as I see, we gain nothing by spawning a single greenlet and waiting for it.

I think the correct way is to split rendering of issues a cross a pool and wait for all of them instead.

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +224,302 @@ def _get_project_status(project_statuses, status):
...
+        Args:
+            path (str): The path of exported markdown file.
+            
+        """
+        text = f"""
+### All User Stories \n
+"""
+        all_stories_template = self.__get_stories_required_data()
+
+        for story in all_stories_template:
```
abom, line 507: I still think using a template **file** is better than building this markdown like this.

OmarElawady:
The hash method is needed, don't know for what though but it raises an exception without it. It's called with this backtrace:
```
-> user_id = self._get_user_id(username)
> /home/omar/codescalers/js-sdk/jumpscale/clients/taiga/taiga.py(29)__hash__()
-> return hash(str(self))
```

jumpscale/clients/taiga/taiga.py
```diff
@@ -342,23 +342,26 @@ def export_all_issues_details(self, path="/tmp/issues_details.md", with_descript
...
         """
-        all_projects_template = list()
-        for project in projects:
-            single_project_template = dict()
-            single_project_template["name"] = project.name
-            single_project_template["created_date"] = project.created_date
-            single_project_template["owner"] = project.owner.get("full_name_display")
-            single_project_template["issues"] = project.list_issues()
-            all_projects_template.append(single_project_template)
+        threads = [gevent.spawn(self.__get_single_project_required_data, project) for project in projects]
```
xmonader, line 362: please name to greenlets instead of threads

jumpscale/clients/taiga/taiga.py
```diff
@@ -342,23 +342,26 @@ def export_all_issues_details(self, path="/tmp/issues_details.md", with_descript
...
+        gevent.joinall(threads)
+        all_projects_template = [thread.value for thread in threads]
```
xmonader, line 364: what happens in case of errors here?

OmarElawady: Very good question. It returns None for the failed ones. Should I filter out the failed ones, raise an exception in case any one of them failed, retry for 6 times for all of them to succeed and raise in case all of the tries failed?

xmonader: No need to raise (join should have raise_exceptions or similar flag), you can link greenlet to link_exception and maybe have a log error there in that handling, or use a group and wrap the errors e.g https://gist.github.com/progrium/956006

jumpscale/clients/taiga/taiga.py
```diff
@@ -217,3 +224,306 @@ def _get_project_status(project_statuses, status):
...
+            projects(list): list of projects
+
+        Returns:
+            all_projects_template (dic): dict of the data required from project object
+
+        """
+        greenlets = [gevent.spawn(self.__get_single_project_required_data, project) for project in projects]
+        gevent.joinall(greenlets)
+        all_projects_template = [
+            greenlet.value for greenlet in greenlets if greenlet.value
```
xmonader, line 365: what if the value of the greenlet itself is None? I think better to check on .successful instead of .value

xmonader (approved):

waleedhammam (approved):

PR #1411, opened 2020-10-12
check if the owner is the same as the pool owner

OmarElawady:
### Description

The gateway pool is identified only by the farm name. This causes issues when switching identities, since the pool owner changes and the gateways available on mainnet are different from testnet. In this PR, a check is added for the owner and a new pool is created in case of mismatch.

Screenshot from 2020-10-12 12-48-42


waleedhammam:
We agreed to add the owner/explorer in the model to be easy in comparision and avoid duplication

m-motawea (approved):

xmonader (approved):

PR #1430, opened 2020-10-14
do hash the shared secret generated to decrypt workloads sensitive in formation

zaibon:
Extract from https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman#Key_establishment_protocol:
> While the shared secret may be used directly as a key, it can be desirable to hash the secret to remove weak bits due to the Diffie–Hellman exchange.

DO NOT MERGE YET.

### DIFF
--- a/jumpscale/sals/zos/crypto.py
+++ b/jumpscale/sals/zos/crypto.py
@@ -1,18 +1,20 @@
 import binascii
+from hashlib import blake2b
 from typing import Union
 
 from jumpscale.core.identity import Identity
 from jumpscale.loader import j
 from nacl.bindings import crypto_scalarmult
-from nacl.public import SealedBox
+from nacl.secret import SecretBox
 from nacl.signing import VerifyKey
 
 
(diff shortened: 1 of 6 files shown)


m-motawea (approved):

waleedhammam (approved):

jumpscale/sals/zos/crypto.py
```diff
@@ -1,11 +1,15 @@
...
-from nacl.public import SealedBox
+from jumpscale.core.identity import Identity
+from jumpscale.loader import j
+from nacl.bindings import crypto_scalarmult
+from nacl.secret import SecretBox
 from nacl.signing import VerifyKey
 
 
-def encrypt_for_node(public_key: str, payload: Union[str, bytes]) -> str:
+def encrypt_for_node(identity: Identity, public_key: str, payload: Union[str, bytes]) -> str:
```
xmonader, line 12: Can we please update the description, because we don't use only the node public key anymore correct?

xmonader (approved):

PR #1458, opened 2020-10-19
select the first available farm and gateway

m-motawea:
### Description

speed up farm and gateway selection in apps chatflow by using the first available object.

### DIFF
--- a/jumpscale/sals/marketplace/apps_chatflow.py
+++ b/jumpscale/sals/marketplace/apps_chatflow.py
@@ -43,7 +43,7 @@ class MarketPlaceAppsChatflow(MarketPlaceChatflow):
         self.solution_metadata["owner"] = self.username
         self.threebot_name = j.data.text.removesuffix(self.username, ".3bot")
         self.ip_version = "IPv6"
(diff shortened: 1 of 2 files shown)


jumpscale/sals/marketplace/apps_chatflow.py
```diff
@@ -74,9 +74,10 @@ def _choose_flavor(self, flavors=None):
 
     def _get_pool(self):
         self.currency = "TFT"
-        available_farms = []
         farm_names = [f.name for f in j.sals.zos.get()._explorer.farms.list()]
+        random.shuffle(farm_names)
```
waleedhammam, line 78: what about instead of shuffle, we sort the largest resources first to reduce the probability of no ssd storage in user pools, as in apps ?

m-motawea: I think shuffling is better to avoid getting stuck with a mal-functioning farm for all deployments

waleedhammam (approved):

xmonader (approved):

Issue #1314, opened 2020-09-29
Test Coverage For Github


mhost39:

Dina-Abd-Elrahman:
**Commit_ID** f6605c7c855ae3cfa036fde06dab6b7b58bb03aa, **Branch** development

![ff](https://user-images.githubusercontent.com/21218053/97105609-d81b4900-16c4-11eb-9c97-826584cd45b6.png)

PR #1459, opened 2020-10-19
add wait for create repo

mhost39:
### Description
Test coverage for Github

### Related Issues
https://github.com/threefoldtech/js-sdk/issues/1314

### DIFF
--- a/tests/clients/github/test_github.py
+++ b/tests/clients/github/test_github.py
@@ -30,7 +30,7 @@ class GithubClientTest(BaseTests):
             j.sals.fs.rmtree(f"/tmp/{self.directory_name}/")
 
     def wait_for_deleting_repo(self, sec, repo):
-        while sec:
+        for i in range(sec):
             if repo.name in str(self.client.get_repos()):
                 sleep(1)
                 sec -= 1
@@ -39,7 +39,7 @@ class GithubClientTest(BaseTests):
         return False
 
     def wait_for_creating_repo(self, sec, repo):
-        while sec:
+        for i in range(sec):
             if repo.name not in str(self.client.get_repos()):
                 sleep(1)
                 sec -= 1
--- a/tests/clients/github/test_github.py
+++ b/tests/clients/github/test_github.py
@@ -30,7 +30,7 @@ class GithubClientTest(BaseTests):
             j.sals.fs.rmtree(f"/tmp/{self.directory_name}/")
 
     def wait_for_deleting_repo(self, sec, repo):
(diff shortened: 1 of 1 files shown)


tests/clients/github/test_github.py
```diff
@@ -31,7 +31,16 @@ def tearDown(self):
 
     def wait_for_deleting_repo(self, sec, repo):
         while sec:
-            if repo.full_name in str(self.client.get_repos()):
+            if repo.name in str(self.client.get_repos()):
+                sleep(1)
+                sec -= 1
```
waleedhammam, line 36: What is sec ? A counter ? Maybe we can replace while with for then

tests/clients/github/test_github.py
```diff
@@ -31,7 +31,16 @@ def tearDown(self):
...
+            else:
+                return True
+        return False
+
+    def wait_for_creating_repo(self, sec, repo):
+        while sec:
+            if repo.name not in str(self.client.get_repos()):
```
waleedhammam, line 43: Here too

tests/clients/github/test_github.py
waleedhammam: nope, that's not the correct for for loop :)

waleedhammam (changes requested):

tests/clients/github/test_github.py
```diff
@@ -30,10 +30,17 @@ def tearDown(self):
             j.sals.fs.rmtree(f"/tmp/{self.directory_name}/")
 
     def wait_for_deleting_repo(self, sec, repo):
-        while sec:
-            if repo.full_name in str(self.client.get_repos()):
+        for i in range(sec):
```
AhmedHanafy725, line 33: In case of we don't need the counter variable, it's better to use `_` instead.
```suggestion
        for _ in range(sec):
```

tests/clients/github/test_github.py
```diff
@@ -30,10 +30,17 @@ def tearDown(self):
...
+            if repo.name in str(self.client.get_repos()):
+                sleep(1)
+            else:
+                return True
+        return False
+
+    def wait_for_creating_repo(self, sec, repo):
+        for i in range(sec):
```
AhmedHanafy725, line 41:
```suggestion
        for _ in range(sec):
```

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

waleedhammam (approved):

Issue #1018, opened 2020-09-06
Check wallet balance before doing the transaction for creating/extending a pool


Dina-Abd-Elrahman:
I can choose a wallet while extending the pool even if this wallet is empty. 

![ff](https://user-images.githubusercontent.com/21218053/92329980-cd044f00-f06b-11ea-8fd5-978782c4f301.png)
![ff1](https://user-images.githubusercontent.com/21218053/92329982-ce357c00-f06b-11ea-88fe-7ff719f735a6.png)
![image](https://user-images.githubusercontent.com/41957921/92330383-cc20ec80-f06e-11ea-86fa-b124cc64d404.png)

xmonader:
fillter out the wallets without enough balance

Dina-Abd-Elrahman:
**Commit_ID** 8996cfa597aecfa66fb8c405779c43e921da6e3e, **Branch** development

![ff](https://user-images.githubusercontent.com/21218053/96888781-432b0c80-1486-11eb-8791-63f77d77af15.png)

![ff](https://user-images.githubusercontent.com/21218053/96889191-a87efd80-1486-11eb-9175-566339421f49.png)

PR #1488, opened 2020-10-22
Show wallets with enough balance when paying for pools

ranatrk:
### Description
When paying for pool, wallets with not enough balance appear in the options and so fail if chosen

### Changes

Check for wallet balance against the amount to be payed for the pool and filter out wallets with enough balance

- no wallets with enough balance:
Screenshot from 2020-10-22 12-32-52

- wallet with enough balance added:
Screenshot from 2020-10-22 12-23-26

### Related Issues

https://github.com/threefoldtech/js-sdk/issues/1018

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/jumpscale/sals/reservation_chatflow/deployer.py
+++ b/jumpscale/sals/reservation_chatflow/deployer.py
@@ -322,7 +322,7 @@ class ChatflowDeployer:
         cu = form.int_ask("Required Amount of Compute Unit (CU)", required=True, min=0, default=0)
         su = form.int_ask("Required Amount of Storage Unit (SU)", required=True, min=0, default=0)
         time_unit = form.drop_down_choice(
-            "Please choose the duration unit", ["Day", "Month", "Year"], required=True, default="Month",
+            "Please choose the duration unit", ["Day", "Month", "Year"], required=True, default="Month"
         )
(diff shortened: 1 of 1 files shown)


jumpscale/sals/reservation_chatflow/deployer.py
```diff
@@ -477,7 +477,17 @@ def show_payment(self, pool, bot):
...
         wallet_names = []
         for w in wallets.keys():
-            wallet_names.append(w)
+            wallet = j.clients.stellar.get(w)
+            try:
+                balances = wallet.get_balance().balances
+            except:
+                continue
+            for balance in balances:
+                if balance.asset_code in escrow_asset:
```
xmonader, line 486: can join the two conditions here in one 

```suggestion
                if balance.asset_code in escrow_asset and balance.balance > total_amount :
```

ranatrk: @xmonader  I did it separately to break if the balance isnt enough but keep iterating in the balances if the asset didnt match yet. if they are combined then we'll remove the break and it will iterate even if the balance wasn't matching. Should that matter?

xmonader (approved):

waleedhammam (approved):

PR #1492, opened 2020-10-22
add new tools to compute how much cloud units is used by a workload

### LINKED ISSUE #1423: Creation of a tool to compute how much cloud unit a workload consumes
It would be useful to have a library that allows user to know how much a workload will consume in term of cloud units.
This allow user to know how much capacity he needs to have in its pool before provisioning the workloads.

The code to compute how much resource units a workload comsume is available in the explorer:
- container: https://github.com/threefoldtech/tfexplorer/blob/046188aa3f9383fdb5e17f420e6d62f5fb4ba18a/models/generated/workloads/container.go#L119
- k8s: https://github.com/threefoldtech/tfexplorer/blob/master/models/generated/workloads/k8s.go#L24
volume: https://github.com/threefoldtech/tfexplorer/blob/master/models/generated/workloads/volume.go#L18
- zdb: https://github.com/threefoldtech/tfexplorer/blob/master/models/generated/workloads/zdb.go#L22

zaibon:
### Description

Add a new `consumption` section in `j.tools.zos` which allow to compute how much cloud units are consumed by a workload object

### Changes

- Add new module in j.tools.zos
- Add `resource_units` methods on all the workloads model

### Related Issues

fixes #1423

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/jumpscale/clients/explorer/models.py
+++ b/jumpscale/clients/explorer/models.py
@@ -116,7 +116,7 @@ class CloudUnits(Base):
 
 
 class ResourceUnitAmount(Base):
-    cru = fields.Float()
+    cru = fields.Integer()
     mru = fields.Float()
     hru = fields.Float()
     sru = fields.Float()
@@ -491,7 +491,7 @@ class Container(Base):
         cap = self.capacity
         resource_units = ResourceUnitAmount()
         resource_units.cru = cap.cpu
-        resource_units.mru = round(cap.memory / 1024)
+        resource_units.mru = round(cap.memory / 1024 * 10000) / 10000
         storage_size = round(cap.disk_size / 1024 * 10000) / 10000
         storage_size = max(0, storage_size - 50)  # we offer the 50 first GB of storage for container root filesystem
         if cap.disk_type == DiskType.HDD:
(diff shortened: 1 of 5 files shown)


jumpscale/clients/explorer/test_models.py
```diff
@@ -0,0 +1,80 @@
+from .models import DiskType, Container, Volume, ZdbNamespace, K8s
```
xmonader, line 1: can we move the tests into tests/clients/explorer instead

jumpscale/clients/explorer/models.py
```diff
@@ -412,6 +471,19 @@ class Container(Base):
     capacity = fields.Object(ContainerCapacity)
     info = fields.Object(ReservationInfo)
 
+    def resource_units(self):
+        cap = self.capacity
+        resource_units = ResourceUnitAmount()
+        resource_units.cru = cap.cpu
+        resource_units.mru = round(cap.memory / 1024)
```
LeeSmet, line 494: mru is a float so rounding here only makes sense if we want to limit the amount of significant digits like it is done on the next line imo

LeeSmet (changes requested):
I don't really like the idea of giving 50GB of SSD for containers for free, but it is what it is I guess

LeeSmet (approved):
lgtm

xmonader (approved):

xmonader (approved):

PR #1522, opened 2020-10-26
Fix for threebot packages tests to handle paths

waleedhammam:
### Description

Fix for threebot packages tests to handle paths

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/tests/servers/threebot/test_threebot.py
+++ b/tests/servers/threebot/test_threebot.py
@@ -162,13 +162,16 @@ class Test3BotServer(BaseTests):
         - Try to delete non exists package, and make sure that the error has been raised.
         """
         self.info("Add a package")
+        from jumpscale.packages import marketplace
+        path = j.sals.fs.dirname(marketplace.__file__)
+
         marketplace = j.servers.threebot.default.packages.add(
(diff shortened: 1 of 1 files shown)


jumpscale/servers/threebot/threebot.py
```diff
@@ -268,11 +269,15 @@ def bottle_servers(self):
 
     @property
     def actors(self):
-        return self._load_files(self.actors_dir)
+        if self.actors_dir:
+            return self._load_files(self.actors_dir)
```
abom, line 273: I think the return value is not consistent here as `self._load_files` would return a generator here?

Also, IMO, we can do this change in `self._load_files` instead of repeating it in every method which calls `_load_files`.

jumpscale/servers/threebot/threebot.py
```diff
@@ -216,6 +216,7 @@ def _load_files(self, dir_path):
                 yield dict(name=name, path=file_path)
 
     def load_config(self):
+        print(">>>>", self.path)
```
abom, line 219: debug print?

abom (changes requested):

ranatrk (approved):

abom (approved):

AhmedHanafy725 (approved):

Issue #1445, opened 2020-10-17
Use "Display name" instead of "Identity name" and add a tooltip


xmonader:
![image](https://user-images.githubusercontent.com/64129/96339845-5c534800-1097-11eb-8308-5649e4362e8c.png)

As identity name is used within js config, it should be a valid python identifier.
and there's also a typo in the error message (om instead of on)

- should be called `Display name`
- should be unique
- should be valid python identifier
- Should have a tooltip next to it

waleedhammam:
Should be a valid python identifier aswell

mhost39:
branch: development
commit: 8e343adb664db071a46153e0c0b3f31e0376aef5
![image](https://user-images.githubusercontent.com/31102305/98223867-61e7d380-1f5b-11eb-83af-5e97c9750fe7.png)

PR #1531, opened 2020-10-27
use display name only

OmarElawady:
### Description

Replace instance name and tname with display_name 
### Changes

1. Create identity with instance name "{display_name}_{network used}" to allow registering the same tname on testnet and mainnet.
2. Tooltip with message "The name to be registered on TFGrid, used to reference your created 3Bot identity.". 

### Related Issues

https://github.com/threefoldtech/js-sdk/issues/1445


waleedhammam (approved):

xmonader:
I don't think it should be done this way, I believe the display name in the UI part get translated to identity_name in the backend without appending the network type

xmonader (approved):

PR #1542, opened 2020-10-27
Separate mainnet wallet from identity

OmarElawady:
### Description

The mainnet wallet network was being determined by the default identity, which incorrectly assumed that is always a mainnet identity. Now it should be independent.

### DIFF
--- a/jumpscale/packages/admin/actors/wallet.py
+++ b/jumpscale/packages/admin/actors/wallet.py
@@ -5,11 +5,13 @@ from jumpscale.clients.stellar.stellar import _NETWORK_KNOWN_TRUSTS
 
 class Wallet(BaseActor):
     @actor_method
-    def create_wallet(self, name: str) -> str:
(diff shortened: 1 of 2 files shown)


ranatrk (approved):

jumpscale/packages/admin/actors/wallet.py
```diff
@@ -5,11 +5,13 @@
 
 class Wallet(BaseActor):
     @actor_method
-    def create_wallet(self, name: str) -> str:
+    def create_wallet(self, name: str, wallettype: str = None) -> str:
         explorer = j.core.identity.me.explorer
-        wallettype = "STD"
-        if "testnet" in explorer.url or "devnet" in explorer.url:
-            wallettype = "TEST"
+        if wallettype is None:
```
waleedhammam, line 10: maybe we can do 
```suggestion
        if not wallettype:
```

Issue #1578, opened 2020-11-02
Update Example Scripts to use the management network or create it if it doesn't exist


m-motawea:
### Description

Currently all example scripts create a new network and use it for deployment. 
we need to update these scripts to use the management network if it exists or create it if it doesn't with subnet `10.100.0.0/16`

### Scripts:
- [x] k8s (@m-motawea & @waleedhammam)
- [x] minio (@m-motawea & @waleedhammam)
- [x] ubuntu (@RafyAmgadBenjamin )
- [x] ubuntu deploy (@AhmedSa-mir & @dmahmouali )
- [x] ubuntu server (@m-motawea )

RafyAmgadBenjamin:
Ubuntu PR:
https://github.com/threefoldtech/js-sdk/pull/1590

PR #1590, opened 2020-11-02
Update ubuntu deploy script to use management network with the new ip

RafyAmgadBenjamin:
### Description

Update ubuntu script to use `management` network with the correct IP, if we are not able to load the network then we create a new one.
### Changes

List of changes this PR includes
examplescripts/ubuntu.py

### Related Issues

List of related issues
https://github.com/threefoldtech/js-sdk/issues/1578

image

### DIFF
--- a/examplescripts/ubuntu.py
+++ b/examplescripts/ubuntu.py
@@ -2,7 +2,7 @@ from jumpscale.loader import j
 from time import sleep
 import uuid
 
-zos = j.sals.zos
+zos = j.sals.zos.get()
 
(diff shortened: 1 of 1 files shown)


ashraffouda (approved):

m-motawea:
you need to check if the selected node is already part of the network or not. if it isn't, you need to find a free range and add it to the network. if it is, you need to find a free ip address from that node's range and use it for deployment

m-motawea (changes requested):
you need to check if the selected node is already part of the network or not. if it isn't, you need to find a free range and add it to the network. if it is, you need to find a free ip address from that node's range and use it for deployment

m-motawea (approved):

PR #1605, opened 2020-11-03
update ubuntu_deploy script

dmahmouali:
### Related Issues

[Update Example Scripts to use the management network or create it if it doesn't exist](https://github.com/threefoldtech/js-sdk/issues/1578)


examplescripts/ubuntu_deploy.py
```diff
@@ -1,123 +1,202 @@
...
+    return False
+
+
+def get_node_free_ip(ip_range):
+    """
+    get a node free ip, skips the first ip as it is assigned to the network
+    """
+    hosts = netaddr.IPNetwork(ip_range).iter_hosts()
+    next(hosts)
+    return str(next(hosts))  # skip ip used by node
```
m-motawea, line 42: this doesn't really get a free IP. it will always return `10.100.x.2`. to get a free IP you need to check all workloads deployed on the node and attached to the target network and exclude their IPs from the node range.

examplescripts/ubuntu_deploy.py
```diff
@@ -1,123 +1,202 @@
...
+        break
+else:
+    raise Exception("can not find a node with ipv4 that can be used as access node for the cluster")
+
+freerange1 = get_free_iprange(ip_range, network_used_ranges)
+zos.network.add_node(network, access_node.node_id, freerange1, pool.pool_id)
+network_used_ranges.add(freerange1)
+
+freerange2 = get_free_iprange(ip_range, network_used_ranges)
+wg_quick = zos.network.add_access(network, access_node.node_id, freerange2, ipv4=True)
```
m-motawea, line 121: why are you always adding access to the network? if the network is already created, there is no need to add access.

m-motawea (changes requested):

examplescripts/ubuntu_deploy.py
```diff
@@ -1,130 +1,271 @@
...
+
+
+zos = j.sals.zos.get()
+WALLET_NAME = "test"
+network_name = "management"
+ip_range = "10.100.0.0/16"
+wg_quick = None
+
+
+bad_nodes = [
```
xmonader, line 24: is this still needed?

m-motawea: there are still some nodes that fail because of the storage issue "not enough capacity in pool of type ssd" and there is one node that is not responding to any workload

waleedhammam (approved):

xmonader (approved):

m-motawea (approved):

Issue #1584, opened 2020-11-02
3bots activate the testnet wallet using friendbot, not the testnet activation service


robvanmieghem:
### Description

Although this works now and prevents our testnet activation service from being drained, this does not test the actual activation service/flow itself.
It also does not test real life scenario's where the activation service is in need of new funding and operations needs to step in

robvanmieghem:
deprecated

PR #1709, opened 2020-11-19
Activate threebot testnet wallets through the testnet activation service

robvanmieghem:
Fixes issue #1584 
I also moved the funding of the testnet 3bot wallet to the 3bot code since it belongs there, it's not generic wallet functionality.

### DIFF
--- a/jumpscale/clients/stellar/__init__.py
+++ b/jumpscale/clients/stellar/__init__.py
@@ -1,5 +1,3 @@
-import requests
-
 from jumpscale.loader import j
 from jumpscale.core.base import StoredFactory
 from stellar_sdk import Keypair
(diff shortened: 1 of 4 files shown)


jumpscale/clients/stellar/__init__.py
```diff
@@ -51,45 +49,6 @@ def check_stellar_service(self):
 
         return services_status
 
-    def create_testnet_funded_wallet(self, name: str) -> bool:
```
waleedhammam, line 54: I think this method is used a lot, for example in admin dashboard and the unit tests, changing it's location/signature may lead to break these things ?

jumpscale/packages/admin/actors/wallet.py
```diff
@@ -19,15 +19,11 @@ def create_wallet(self, name: str, wallettype: str = None) -> str:
         wallet = j.clients.stellar.new(name=name, network=wallettype)
 
         try:
-            if wallettype == "TEST":
-                wallet.activate_through_friendbot()
-            else:
-                wallet.activate_through_threefold_service()
+            wallet.activate_through_threefold_service()
```
waleedhammam, line 22: activating thought testnet using threefold service activates but gives me that error, is that related ?


```
TypeError: string indices must be integers
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/waleed/js-sdk/jumpscale/clients/stellar/stellar.py", line 246, in activate_through_threefold_service
    self._activation_account(activationdata["activation_code"])
    │                        └ '{"activation_code": "abcd", "address": "GAJ4FG4DA2M3BUECSANIRX3NWVVHICUBMS2Y24Z4ECLCMZFZCDE6T3YM", "phonenumbers": ["+123456789...
    └ Stellar(
  instance_name='ell',
  network=<Network.TEST: 'TEST'>,
  address='GAJ4FG4DA2M3BUECSANIRX3NWVVHICUBMS2Y24Z4ECLCMZFZCDE...
TypeError: string indices must be integers

string indices must be integers
JS-NG> wallet.get_balance()                                                                                                                                                                                   
Balances
  3.6000000 XLM
```

jumpscale/clients/stellar/__init__.py
robvanmieghem: The admin dashboard uses the wallet actor to create a new account and only the threebot should fund its testnet wallet. If this breaks a test, we should revise the test.

jumpscale/packages/admin/actors/wallet.py
robvanmieghem: The activation service was rewritten  but this is exactly why testnet should not use a different setup than mainnet, the activation service on testnet is not tested at the moment. Good thing you found this, will check what is going wrong

robvanmieghem: Found it, there is an avtivateaccount call that does nothing and should not be called, It used to return no error but it does in the rewrite. Will fix it.

robvanmieghem: Done, added the commit to this PR

robvanmieghem: Apparently no-one noticed the error before even though it's a client error and is in there since april or so

waleedhammam: thanks

waleedhammam:
@robvanmieghem I think we can close this now as testnet wallets/tokens aren't used anymore

robvanmieghem:
@waleedhammam  indeed

Issue #1786, opened 2020-12-04
Add poetry build to the workflow


xmonader:
projects needs to always pip installable with correct dependencies, to make sure of that let's add poetry build to the workflow

sameh-farouk:
there are many workflows, I assume you mean the js-sdk.yml one?

sameh-farouk:
also since pip 19.0, it becomes possible to pip install a Poetry-managed Python package from a git repository without setup.py
ex: `pip install git+{git-repository-url}`
so I guess we only need the output of the `poetry build` command if we want to publish it `poetry publish` to an external repository (likely PyPi)?

PR #1792, opened 2020-12-08
add poetry build to workflow

mhost39:
### Description
add poetry build to the workflow

### Related Issues
https://github.com/threefoldtech/js-sdk/issues/1786#event-4074702792

### DIFF
--- a/.github/workflows/js-sdk.yml
+++ b/.github/workflows/js-sdk.yml
@@ -45,7 +45,8 @@ jobs:
           poetry build
       - name: Extract tar files
         run: |
-          tar xvzf dist/js-sdk-11.1a1.tar.gz
+          tar xvzf dist/js-sdk-*.tar.gz -C /tmp
       - name: Install
         run: |
-          sudo python3 js-sdk-11.1a1/setup.py install
+          cd /tmp/js-sdk-*
+          sudo python3 setup.py install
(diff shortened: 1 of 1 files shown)


.github/workflows/js-sdk.yml
```diff
@@ -40,3 +40,12 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Build package
+        run: |
+          poetry build
+      - name: Extract tar files
+        run: |
+          tar xvzf dist/js-sdk-11.1a1.tar.gz
```
xmonader, line 48: this will break if we ever update the version in pyproject.toml

.github/workflows/js-sdk.yml
```diff
@@ -40,3 +40,13 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Build package
+        run: |
+          poetry build
+      - name: Extract tar files
+        run: |
+          tar xvzf dist/js-sdk-*.tar.gz -C /tmp
```
waleedhammam, line 48: using '*' is so confusing and may prone errors

xmonader:
should use

Issue #2059, opened 2020-12-30
Prepare 3Bot deployer/server for STD wallet migration


waleedhammam:
### Description

https://github.com/threefoldtech/js-sdk/blob/646678a96c3027bafaf51e85967242bd60f05016/jumpscale%2Fentry_points%2Fthreebot.py#L277

should be STD only

- Also in threebot deployer package.py

https://github.com/threefoldtech/js-sdk/blob/5341abc657ee5958480f4800346727df06531179/jumpscale%2Fpackages%2Fthreebot_deployer%2Fpackage.py#L12

PR #2065, opened 2020-12-30
migrate threebot deployer and threebot server to use STD wallet

waleedhammam:
### Description

- Limit 3bot deployer to only use STD wallets
- Limit 3bot deployer demos wallet to be STD

### Related Issues

- https://github.com/threefoldtech/js-sdk/issues/2059

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/jumpscale/entry_points/threebot.py
+++ b/jumpscale/entry_points/threebot.py
@@ -248,21 +248,11 @@ def cli():
 
 def have_wallets():
     wallets = j.clients.stellar.list_all()
-    test, main = False, False
     for wallet_name in wallets:
         wallet = j.clients.stellar.get(wallet_name)
-        if wallet.network.value == "TEST":
-            test = True
-        elif wallet.network.value == "STD":
-            main = True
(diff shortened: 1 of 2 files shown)


jumpscale/entry_points/threebot.py
```diff
@@ -248,21 +248,9 @@ def cli():
...
-            main = True
-    return test, main
-
-
-def create_test_wallet(wallet_name):
-    try:
-        j.clients.stellar.create_testnet_funded_wallet(wallet_name)
-    except Exception as e:
-        j.logger.error(str(e))
+        return wallet.network.value == "STD"
```
xmonader, line 253: this is an early return .. should return True in case of the first one, but also not return False immediately before exhausting the loop

xmonader (changes requested):
Requested changes

xmonader (approved):

ranatrk (approved):

PR #2128, opened 2021-01-05
Check Check if the vdc is already created to avoid minimal entrypoint re-rerun failure

OmarElawady:
### Description

- Modify a copy of the passed dict in vdc `from_dict` method.
- Check if the vdc is already created to avoid entrypoint re-rerun failure.


dmahmouali (approved):
Make sure to change `SDK_VERSION` before the merge.

xmonader (approved):

Issue #2151, opened 2021-01-09
update pool payment and extension checks for capacitypool using the get payment info


xmonader:
get payment info api got added on devnet, testnet. let's prepare the changes needed for better checks on pools extensions. code already exists on development_vdc

waleedhammam:
Verified

PR #2156, opened 2021-01-10
wait for pool payments

m-motawea:
### Related Issues

https://github.com/threefoldtech/js-sdk/issues/2151

### DIFF
--- a/jumpscale/clients/explorer/models.py
+++ b/jumpscale/clients/explorer/models.py
@@ -665,3 +665,16 @@ class PoolEscrow(Base):
 class PoolCreated(Base):
     reservation_id = fields.Integer()
     escrow_information = fields.Object(PoolEscrow)
+
+
+class PoolPayment(Base):
+    id = fields.Integer()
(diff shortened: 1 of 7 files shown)


jumpscale/clients/explorer/models.py
```diff
@@ -665,3 +665,16 @@ class PoolEscrow(Base):
...
     escrow_information = fields.Object(PoolEscrow)
+
+
+class PoolPayment(Base):
+    id = fields.Integer()
+    farmer_id = fields.Integer()
+    address = fields.String()
+    expiration = fields.DateTime()
+    asset = fields.String()
+    amount = fields.Integer()
```
waleedhammam, line 676: Amount can be float ?

ranatrk (approved):

jumpscale/clients/explorer/models.py
m-motawea: it is the same as returned by the reservation escrow which is integer

waleedhammam (approved):

PR #2165, opened 2021-01-10
Development vdc scripts

m-motawea:
1- `init.sh`: configures dhcp and nat (run only once)
2- `setup_vdc.sh`: configures a network namepsace and starts wireguard inside of it
3- `clean_vdc.sh`: deletes the network namespaces and the associated interfaces
4- `connect_vdc.sh`: ssh into the specified ip from the vdc namespace


ashraffouda:
is it better to put those scripts under scripts/network? maybe we have other scripts in the future or so?

Issue #1672, opened 2020-11-11
Automate chatflows tests are failing


AhmedHanafy725:
Please check the tests https://github.com/threefoldtech/js-sdk/runs/1382823533?check_suite_focus=true
- If the failure in (test, gedispatch, or workflow), please fix it.
- If the failure in chatflow, please report the issue and skip the test.

mhost39:
related issue https://github.com/threefoldtech/js-sdk/issues/1678

mhost39:
https://github.com/threefoldtech/js-sdk/runs/1415085572?check_suite_focus=true

sometime when deploy solution get low resource error
and sometime solution is deployed but can't access it like (k8s, minio, ubuntu)
and threeboy work fine in local machine but in CI git this error

![image](https://user-images.githubusercontent.com/31102305/99547004-b65d6b00-29bf-11eb-90d8-6974b3a4de99.png)

m-motawea:
@mhost39 the attribute error in the screenshot is because threebot_deployer package is not installed.
you need  to install the package before you run the test

AhmedHanafy725:
This error happens with cert check even with setting `TEST_CERT`
![image](https://user-images.githubusercontent.com/41957921/99903434-08a6d080-2ccd-11eb-9360-fe3dad6058b2.png)

mhost39:
https://github.com/threefoldtech/js-sdk/issues/1945

PR #2181, opened 2021-01-11
Fix chatflows tests

mhost39:
### Related Issues
https://github.com/threefoldtech/js-sdk/issues/1672


AhmedHanafy725 (changes requested):
some tests wait with `timeout` and the other wait with `deployment_timeout` and both are 60 sec if both are the same, please unify

AhmedHanafy725 (approved):

ramezsaeed (approved):

Issue #2242, opened 2021-01-14
The deployment message and initialization check are tightly coupled with a specific solution type


OmarElawady:
### Description

The solution assumes that there's a domain name assigned in the initialization step to check if it's reachable. Which is not the case for the blockchain solution

xmonader:
maybe better to define REQUIRES_DOMAIN = True as the default and for solutions that don't have domain we set it to False and skip the whole flow for domains

MElborolossy:
Verified
Digibyte (Blockchain solution) deployed successfully
![image](https://user-images.githubusercontent.com/11272864/109139385-25d3bb00-7764-11eb-95f8-5349db2661ed.png)

PR #2259, opened 2021-01-14
ignore domain operations if the domain step is not executed

OmarElawady:
### Description

ignore domain operations if the domain step is not executed

Tested on taiga, mattermost, and a domainless solution.

### Related issues

https://github.com/threefoldtech/js-sdk/issues/2242

### DIFF
--- a/jumpscale/packages/vdc_dashboard/sals/solutions_chatflow.py
+++ b/jumpscale/packages/vdc_dashboard/sals/solutions_chatflow.py
@@ -321,6 +321,9 @@ class SolutionsChatflowDeploy(GedisChatBot):
             return False
         return True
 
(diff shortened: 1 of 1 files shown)


dmahmouali:
We need also to update the readme file to make the `initialization` step not required.
https://github.com/threefoldtech/js-sdk/blob/development_vdc/docs/wiki/tutorials/add_marketplacevdc_chatflow.md#L30

OmarElawady:
well, it's used also for checking if the chart failed due to lack of resources, so it still should be present even if there's no domain.

dmahmouali (approved):

waleedhammam (approved):

Issue #2425, opened 2021-01-24
eVDC created config file does no contain the encryption key?


weynandkuijpers:
Trying to use the binary (self compiled from the repo) to store a file:
```
john@RescuedMac:~/opt/code/github/threefoldtech/0-stor_v2/test$ zstor_v2 store -f ./config.json 
Error: "invalid length 0, expected a hex encoded byte slice of length 32 (64 hex chars) for key `encryption.key` at line 8 column 7"
john@RescuedMac:~/opt/code/github/threefoldtech/0-stor_v2/test$ 
```
I have IPv6 connectivity:
```
john@RescuedMac:~/opt/code/github/threefoldtech/0-stor_v2/test$ ping6 2a02:1802:5e:0:9893:d6ff:fe10:4bb6
PING 2a02:1802:5e:0:9893:d6ff:fe10:4bb6(2a02:1802:5e:0:9893:d6ff:fe10:4bb6) 56 data bytes
64 bytes from 2a02:1802:5e:0:9893:d6ff:fe10:4bb6: icmp_seq=1 ttl=54 time=121 ms
64 bytes from 2a02:1802:5e:0:9893:d6ff:fe10:4bb6: icmp_seq=2 ttl=54 time=120 ms
64 bytes from 2a02:1802:5e:0:9893:d6ff:fe10:4bb6: icmp_seq=3 ttl=54 time=119 ms
64 bytes from 2a02:1802:5e:0:9893:d6ff:fe10:4bb6: icmp_seq=4 ttl=54 time=119 ms
^C
```

What am I doing wrong?

Config file:
```
data_shards = 2
parity_shards = 1
redundant_groups = 0
redundant_nodes = 0

[encryption]
algorithm = "AES"
key = ""

[compression]
algorithm = "snappy"

[[groups]]

[[groups.backends]]
address = "[2a02:1802:5e:0:9ca8:d9ff:feeb:d608]:9900"
namespace = "2516-1"
password = "LeeL00+="

[[groups]]

[[groups.backends]]
address = "[2a02:1802:5e:0:9ca8:d9ff:feeb:d608]:9900"
namespace = "2517-1"
password = "LeeL00+="

[[groups]]

[[groups.backends]]
address = "[2a02:1802:5e:0:9893:d6ff:fe10:4bb6]:9900"
namespace = "2518-1"
password = "LeeL00+="

[[groups]]

[[groups.backends]]
[...]

PR #2426, opened 2021-02-01
Add random value as the default for the zstor encryption key

dmahmouali:
### Changes

- Add random value as the default for the zstor encryption key
```
secrets.token_hex([nbytes=None])¶
Return a random text string, in hexadecimal. The string has nbytes random bytes, each byte converted to two hex digits. If nbytes is None or not supplied, a reasonable default is used.
```
### Related Issues

- [https://github.com/threefoldtech/js-sdk/issues/2425](https://github.com/threefoldtech/js-sdk/issues/2425)
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/jumpscale/packages/vdc_dashboard/bottle/deployments.py
+++ b/jumpscale/packages/vdc_dashboard/bottle/deployments.py
@@ -175,6 +175,8 @@ def cancel_deployment():
 @app.route("/api/zstor/config", method="POST")
 @authenticated
 def get_zstor_config():
+    import secrets
+
     vdc = _get_vdc()
     vdc_zdb_monitor = vdc.get_zdb_monitor()
     password = vdc_zdb_monitor.get_password()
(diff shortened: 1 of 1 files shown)


jumpscale/packages/vdc_dashboard/bottle/deployments.py
```diff
@@ -183,7 +185,7 @@ def get_zstor_config():
         "parity_shards": 1,
         "redundant_groups": 0,
         "redundant_nodes": 0,
-        "encryption": {"algorithm": "AES", "key": "",},
+        "encryption": {"algorithm": "AES", "key": secrets.token_hex(32),},
```
m-motawea, line 188: this will generate a new key every time the config is downloaded which will result in corrupt config
you can use zdb password and zfill it to 32

xmonader (approved):

m-motawea (approved):

PR #2435, opened 2021-02-02
Run js-sdk build tests only when a python file changed

### LINKED ISSUE #2256: Run build tests only on PRs changing .py files only
### Description

We should run js-sdk  build tests only on PRs changing .py files only

sameh-farouk:
### Description

Adding the workflow feature requested in issue #2256

### Changes

- Now js-sdk build tests should run only when a python file `*.py` changed somewhere in the project

- update: after the discussion, `.github/workflows/**.yml` as well as `pyproject.toml` and `poetry.lock` files added to paths and the workflow will trigger if one of these files changed too.

## Notes:

- This affects the js-sdk workflow only. the js-sdk-check-wheel workflow will still work on every Push or PR.

### Related Issues

closes #2256

### DIFF
--- a/.github/workflows/js-sdk.yml
+++ b/.github/workflows/js-sdk.yml
@@ -6,8 +6,12 @@ name: js-sdk
 on:
   push:
     branches: [ development ]
+    paths:
+    - '**.py'
   pull_request:
     branches: [ development ]
+    paths:
+    - '**.py'
 
 jobs:
   build:
--- a/.github/workflows/js-sdk.yml
+++ b/.github/workflows/js-sdk.yml
@@ -8,10 +8,16 @@ on:
     branches: [ development ]
     paths:
     - '**.py'
+    - '.github/workflows/**.yml'
+    - 'pyproject.toml'
+    - 'poetry.lock'
   pull_request:
     branches: [ development ]
     paths:
     - '**.py'
+    - '.github/workflows/**.yml'
+    - 'pyproject.toml'
+    - 'poetry.lock'
 
 jobs:
   build:


abom (approved):

waleedhammam (approved):

sameh-farouk:
I think we also need to add `.github/workflows/` too to paths.
so when we bush a commit that changing the workflow itself, the workflow would run.
agree? @abom @waleedhammam

abom:
@sameh-farouk Isn't that the case already? the workflow is regardless of this path filter?

sameh-farouk:
> @sameh-farouk Isn't that the case already? the workflow is regardless of this path filter?

If at least one path matches a pattern in the paths filter, the workflow runs.
'**.py' will trigger the build anytime you push a python file.
`.github/workflows/**` will also trigger the build anytime you push any file (regardless of its type) to this specific path (where .yml files for the workflow exists). also we could make it more explicit `.github/workflows/**.yml`

waleedhammam:
> > @sameh-farouk Isn't that the case already? the workflow is regardless of this path filter?
> 
> If at least one path matches a pattern in the paths filter, the workflow runs.
> `**.py` will trigger the build anytime you push a python file.
> `.github/workflows/**` will also trigger the build anytime you push any file (regardless of its type) to this specific path (where .yml files for the workflow exists). also we could make it more explicit `.github/workflows/**.yml`

No problem 👍

sameh-farouk:
> I think we also need to add `.github/workflows/**` too to paths.
> so when we bush a commit that changing the workflow itself, the workflow would run.
> agree? @abom @waleedhammam

added as well as pyproject.toml and poetry.lock files

AhmedHanafy725 (approved):

Issue #2498, opened 2021-02-09
VDC: Deployed node after creating the vdc has different ssh key


AhmedHanafy725:
### Steps
1- Deploy a vdc.
2- Go to vdc compute nodes
3- Add node

### Expected
the added node has the same key as the master.

### Result
the added node has a different key, although the key already on the threebot container

master's wid: https://explorer.devnet.grid.tf/api/v1/reservations/workloads/26940
worker wid: https://explorer.devnet.grid.tf/api/v1/reservations/workloads/26948
added node's wid: https://explorer.devnet.grid.tf/api/v1/reservations/workloads/26963

RafyAmgadBenjamin:
### Fixed 
### Branch 
`development_vdc_keys`
### Commit
fc2e8a885aec6adecd9df814dd49e38d5b2db46f

hossnys:
verified 
all three workloads have the same ssh-key 
https://explorer.testnet.grid.tf/api/v1/reservations/workloads/11617
https://explorer.testnet.grid.tf/api/v1/reservations/workloads/11624
https://explorer.testnet.grid.tf/api/v1/reservations/workloads/12397

![Screenshot from 2021-02-24 10-52-19](https://user-images.githubusercontent.com/26298009/108976857-a0330b00-7690-11eb-9d93-0e07feebb965.png)

![Screenshot from 2021-02-24 10-52-01](https://user-images.githubusercontent.com/26298009/108976849-9f01de00-7690-11eb-92d1-ecb86fae26ec.png)

![Screenshot from 2021-02-24 10-53-10](https://user-images.githubusercontent.com/26298009/108976861-a1fcce80-7690-11eb-974d-fb80db09de27.png)

![Screenshot from 2021-02-24 11-11-18](https://user-images.githubusercontent.com/26298009/108977289-0fa8fa80-7691-11eb-9eef-02cb49450408.png)

PR #2587, opened 2021-02-21
Deployed node after creating the vdc has different ssh key

RafyAmgadBenjamin:
### Description

Adding node with the same key as the master to be able to ssh the nodes from the master

### Related Issues

https://github.com/threefoldtech/js-sdk/issues/2498

### DIFF
--- a/jumpscale/sals/vdc/deployer.py
+++ b/jumpscale/sals/vdc/deployer.py
@@ -242,7 +242,7 @@ class VDCDeployer:
             reservation_id = pool_info.reservation_id
             return pool.pool_id, reservation_id
         pool_info = self._retry_call(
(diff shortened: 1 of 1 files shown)


waleedhammam (approved):

RafyAmgadBenjamin:
I have tested it already without this PR and the https://github.com/threefoldtech/js-sdk/issues/2498 was solved 

worker node 
image

extended worker node 
image

I can access both from the master node

we no longer need this PR to fix this issue https://github.com/threefoldtech/js-sdk/issues/2498 as this issue was solved by @m-motawea in this  PR https://github.com/threefoldtech/js-sdk/pull/2579/files

PR #2610, opened 2021-02-22
Update etcd chatflow with more info

MElborolossy:
### Description

Update etcd chatflow with more info related to certificate and add additional msg shown only if cert not verified, This message contains another method to use etcd using kubectl.

### Initialize step
image

### In case cert verified
withoutwarning
### In case cert not verified
image

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/jumpscale/packages/vdc_dashboard/chats/etcd.py
+++ b/jumpscale/packages/vdc_dashboard/chats/etcd.py
@@ -4,7 +4,6 @@ from jumpscale.loader import j
 from jumpscale.sals.chatflows.chatflows import chatflow_step
 from jumpscale.packages.vdc_dashboard.sals.solutions_chatflow import SolutionsChatflowDeploy, POD_INITIALIZING_TIMEOUT
 from jumpscale.packages.vdc_dashboard.sals.vdc_dashboard_sals import get_deployments
-from time import time
 
 CHART_LIMITS = {
     "Silver": {"cpu": "100m", "memory": "128Mi", "no_nodes": "1", "volume_size": "2Gi"},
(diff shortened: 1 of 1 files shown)


jumpscale/packages/vdc_dashboard/chats/etcd.py
```diff
@@ -69,17 +70,56 @@ def initializing(self):
             self.k8s_client.delete_deployed_release(self.release_name)
             self.stop(dedent(stop_message))
 
+        self.is_certified = False
+        start_request_time = time()
```
waleedhammam, line 74: maybe we can replace this with to be more readable
```suggestion
        start_request_time = j.data.time.now().timestamp
```

jumpscale/packages/vdc_dashboard/chats/etcd.py
```diff
@@ -69,17 +70,56 @@ def initializing(self):
...
+        request = None
+        j.logger.debug("ETCD Check certificate")
+        while time() - start_request_time <= timeout:
```
waleedhammam, line 77:
```suggestion
        while j.data.time.now().timestamp - start_request_time <= timeout:
```

jumpscale/packages/vdc_dashboard/chats/etcd.py
```diff
@@ -69,17 +70,56 @@ def initializing(self):
...
+                    ),
+                    md=True,
+                )
+                gevent.sleep(10)
+
+        if not self.is_certified:
+            request = j.tools.http.get(f"https://{self.domain}", timeout=timeout, verify=False)
+
+        # Etcd using TCP, always response return this msg "404 page not found"
+        if not "404 page not found" in str(request.content):
```
waleedhammam, line 95: we can check for the status_code instead, to be 404 ?

MElborolossy: It is preferred to let it check with content, as this response will be different if it fails for anything else.

waleedhammam (approved):

PR #2703, opened 2021-03-04
Update timeout in etcd chatflow to be aligned with msgs appear

MElborolossy:
### Description

Update timeout in etcd chatflow to be aligned with msgs appear, change it from 300 sec (5 min) to 180 sec (3 min)

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


jumpscale/packages/vdc_dashboard/chats/extend_kubernetes.py
```diff
@@ -17,7 +17,7 @@ def flavor(self):
         self.vdc_name = list(j.sals.vdc.list_all())[0]
         self.user_info_data = self.user_info()
         self.username = self.user_info_data["username"]
```
AhmedHanafy725, line 19: I think these 2 lines are not needed in this case.

MElborolossy: yes, you are right

AhmedHanafy725 (approved):

PR #2747, opened 2021-03-11
Support setting a dedicated node

muhamadazmy:
### Description

Add a new method to configure 'dedicated' nodes for a single user.

### Changes

add one method to the explorer client

### Related Issues

https://github.com/threefoldtech/tfexplorer/pull/293


waleedhammam (approved):
Can we Run black on the nodes.py file before merging, it's failing in the CI

xmonader:
Dedicated k8s vm is different now size -1 :)

PR #2808, opened 2021-03-22
fix when cursor is empty

DylanVerstraete:


DylanVerstraete:
ah was not aware this was fixed on master already

PR #2915, opened 2021-04-13
Limit long solution names for Digibyte and Presearch

abdulgig:
### Description

Need to limit unnecessarily long solution names by the user because it is used as release_name is in helm charts to distinguish different services in the cluster and at times the helm charts fail to deploy if the names are more than 15 characters

### DIFF
--- a/jumpscale/packages/vdc_dashboard/chats/digibyte.py
+++ b/jumpscale/packages/vdc_dashboard/chats/digibyte.py
@@ -23,9 +23,16 @@ class DigibyteDeploy(SolutionsChatflowDeploy):
 
     def get_release_name(self):
         self._check_uniqueness()
-        super().get_release_name()
-        if len(self.release_name) > 10:
-            raise StopChatFlow("Solution Name should not be more than 10 characters")
+        self._get_vdc_info()
+        message = "Please enter a name for your solution (will be used in listing and deletions in the future and in having a unique url)"
+        releases = [
+            release["name"]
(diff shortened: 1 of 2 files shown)


jumpscale/packages/vdc_dashboard/chats/digibyte.py
```diff
@@ -24,9 +24,10 @@ def _check_uniqueness(self):
     def get_release_name(self):
         self._check_uniqueness()
         super().get_release_name()
+        if len(self.release_name) > 10:
+            raise StopChatFlow("Solution Name should not be more than 10 characters")
```
waleedhammam, line 28: I think it's better to use the internal check and override that step, for this customized solution, we use the max length 20 in your case you can edit it to be 10 like the following 

```
    @chatflow_step(title="Solution Name")
    def get_release_name(self):
        self._get_vdc_info()
        message = "Please enter a name for your solution (will be used in listing and deletions in the future and in having a unique url)"
        releases = [
            release["name"]
            for release in self.k8s_client.list_deployed_releases()
            if release["namespace"].startswith(self.chart_name)
        ]
        self.release_name = self.string_ask(
            message, required=True, is_identifier=True, not_exist=["solution name", releases], md=True, max_length=10
        )
```

abdulgig: Done as requested

waleedhammam (approved):

PR #3070, opened 2021-05-06
move telegram dependency

mhost39:
### Description

move telegram dependency to temp branch

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


waleedhammam (changes requested):
We can't move the dependency without the code, it'll cause import error!

waleedhammam (approved):

Issue #3065, opened 2021-05-06
Drain the kubernetes node before deleting it


AhmedHanafy725:
see https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/

sameh-farouk:
verifying this...

sameh-farouk:
after a discussion with @MElborolossy, we found that the logic to determine which solutions will get deleted and which will be rescheduled can be enhanced so the user could get consistent behavior with the same solutions.

![Screenshot from 2021-05-31 16-21-18](https://user-images.githubusercontent.com/42457449/120221308-787ab280-c23e-11eb-97d7-a82bba1d41af.png)

![Screenshot from 2021-05-31 16-24-16](https://user-images.githubusercontent.com/42457449/120221336-84ff0b00-c23e-11eb-8fa2-febf720b2b14.png)

MElborolossy:
Still Working on verifying the new logic.

MElborolossy:
### Current State
After testing, we found that most solutions use local storage, this will prevent pods from rescheduled on another node.
For now, we will delete all the releases which on/partially on that node before draining and deleting it.
### Related PR
#3181 (Ready and Waiting to reviewed)

PR #3113, opened 2021-05-24
Drain kubernetes node before deleting

MElborolossy:
### Description

Drain Kubernetes node before deleting it

### Changes

- Check for availability of using drain by check the remaining cluster resources and the pods on the node to be drained
- Inform the user if there are pods that can't be deployed again due to lack of resources in the cluster after deleting the node.
-If the user confirms the deletion, delete the releases that need to be deleted and then drain node, and delete it safely

### Related Issues

#3065

### Screenshots
- list of the pods deployed on the node we want to delete, (Taiga, Cryptpad, Peertube)
image

- Delete the worker for UI (Cryptpad can't be redeployed again because of resources limitation)
delete_worker

- Check that all the remaining pods redeployed on other nodes and solutions still running (Taiga, Peertube)
image

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/jumpscale/packages/vdc_dashboard/bottle/api/deployments.py
+++ b/jumpscale/packages/vdc_dashboard/bottle/api/deployments.py
@@ -71,12 +71,20 @@ def get_kubeconfig() -> str:
 def delete_node():
     data = j.data.serializers.json.loads(request.body.read())
     wid = data.get("wid")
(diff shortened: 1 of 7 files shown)


xmonader (approved):

sameh-farouk:
verifying this...

PR #3120, opened 2021-05-26
Spec

sasha-astiadi:
added new spec file on documentation


xmonader:
These aren't specs -not functional nor technical specs-, these are just changelogs (future changelogs). I guess this was opened by mistake here instead of the home repo as changelogs should be tracked in [home repo](https://github.com/threefoldtech/home/tree/master/products) according to the development process we agreed on 

for js-sdk all changelogs are always tracked in git tags and available on the releases page.

Issue #3081, opened 2021-05-10
we are installing poetry on the server in way that Poetry's dependencies cause conflicts with other packages.


sameh-farouk:
while working on an issue related to update vdc dashboard, I found that the way we install poetry on the server is causing some troubles.
-  When you specify a package in the dev-dependencies that is either a dependency of the poetry itself or it depends on one of the poetry dependencies, and you've installed poetry inside an environment that poetry is supposed to manage, poetry will uninstall this mentioned dependency after invoking poetry install --no-dev, making itself unusable in the future.

currently, we install poetry on the server with `pip install poetry` this will install Poetry's dependencies in a way that causes conflicts with other packages.

the recommended way of installing poetry is
```bash
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
```
this runs a custom installer that will install poetry isolated from the rest of our server system by vendorizing its dependencies.

### Steps to reproduce

try to run one of these commands on the server (or locally but you need to make sure you have installed poetry using pip and that you disabled the env creation)
`poetry install --no-dev` --> will remove some of the poetry dependencies make it unusable.
`poetry install --remove-untracked` --> will remove poetry itself.
### Traceback/Logs/Alerts

Traceback/Logs/Alerts if available
```bash
root@zosv2-04:/sandbox/code/github/threefoldtech/js-sdk# poetry show
Traceback (most recent call last):
  File "/usr/local/bin/poetry", line 5, in <module>
    from poetry.console import main
  File "/usr/local/lib/python3.8/dist-packages/poetry/console/__init__.py", line 1, in <module>
    from .application import Application
  File "/usr/local/lib/python3.8/dist-packages/poetry/console/application.py", line 7, in <module>
    from .commands.about import AboutCommand
  File "/usr/local/lib/python3.8/dist-packages/poetry/console/commands/__init__.py", line 4, in <module>
    from .check import CheckCommand
  File "/usr/local/lib/python3.8/dist-packages/poetry/console/commands/check.py", line 2, in <module>
    from poetry.factory import Factory
  File "/usr/local/lib/python3.8/dist-packages/poetry/factory.py", line 18, in <module>
    from .repositories.pypi_repository import PyPiRepository
  File "/usr/local/lib/python3.8/dist-packages/poetry/repositories/pypi_repository.py", line 33, in <module>
    from ..inspection.info import PackageInfo
[...]

PR #3124, opened 2021-05-27
Fix poetry permissions and path

OmarElawady:
### Description

Apparently the permissions in the container is discarded in the flist when started.

### Changes

1. add export to poetry path addition in ~/.bashrc
2. chmod its permission.

### Related Issues

#3081


sameh-farouk:
Can we replace /root with $HOME to be able to accommodate future changes.

Also could we use --no-dev as option for ‘poetry install’.

xmonader (approved):

Issue #3160, opened 2021-06-05
checking pid file of nginx doesn't always reflect its state


xmonader:
![image](https://user-images.githubusercontent.com/64129/120893490-e4309700-c613-11eb-9e14-b6dc56ac361d.png)

![image](https://user-images.githubusercontent.com/64129/120893495-eb57a500-c613-11eb-9827-010cf7df34cb.png)

apparently checking against the `pid` file isn't enough and at some situations it won't be even cleanup correctly, needs to be hardend with checks against nginx ports as well

abom:
As ports are defined in config, checking for them is not the best solution IMHO.
Maybe we could harden this PID file check by checking for the PID itself (if it's already a running process).

waleedhammam:
Nginx works ok now

Issue #3116, opened 2021-05-25
on macOS, threebot server won't start without sudo


sameh-farouk:
### Description

on macOS, threebot server won't start without `sudo`

**This comment has useful debug/trace info also** https://github.com/threefoldtech/js-sdk/issues/3116#issuecomment-848637414

### Version information

* OS:  macOS 10.15.7
* Project commit/version: js-sdk 11.0b4

### Installation method

as documented on our wiki

### Steps to reproduce

1- activate the env

```sh
poetry shell
```

2- start the server with

```sh
threebot start --local --no-cert  
```

### Traceback/Logs/Alerts
```sh
(js-sdk-DbrXJgm3-py3.9) sameh@samehs-mbp js-sdk % threebot start --local --no-cert                   
2021-05-25 17:03:40.799 | WARNING  | jumpscale.entry_points.threebot:create_wallets_if_not_exists:273 - skipping wallets creation, identity isn't configured yet
2021-05-25 17:03:40.936 | INFO     | jumpscale.entry_points.threebot:check_for_bins:34 - ✅ binaries nginx,redis-server,tmux,git required are installed.
2021-05-25 17:03:40.956 | DEBUG    | jumpscale.entry_points.threebot:test_privileged_ports_bind:58 - nginx: the configuration file /var/folders/d0/4n__wn4d6bv2rkd408ckz6s00000gn/T/tmprs2rpv6e syntax is ok
2021-05-25 17:03:40.957 | DEBUG    | jumpscale.entry_points.threebot:test_privileged_ports_bind:58 - nginx: configuration file /var/folders/d0/4n__wn4d6bv2rkd408ckz6s00000gn/T/tmprs2rpv6e test is successful
2021-05-25 17:03:41.036 | INFO     | jumpscale.core.logging.logging:register:152 - Logging from 'jumpscale.servers.gedis' is now bound to 'gedis' app
2021-05-25 17:03:41.075 | WARNING  | jumpscale.entry_points.threebot:create_wallets_if_not_exists:273 - skipping wallets creation, identity isn't configured yet
2021-05-25 17:03:41.148 | DEBUG    | jumpscale.sals.process:get_pids:476 - ignoring : 639 zombie process.
2021-05-25 17:03:41.358 | DEBUG    | jumpscale.sals.process:get_pids:476 - ignoring : 639 zombie process.
......
[...]

PR #3161, opened 2021-06-05
fix issue#3160 

sameh-farouk:
### Description

adding extra checks to make sure of the real state of Nginx when threebot server starts.

### Changes

is_running() now will check if the PID file specified in the Nginx configuration file Is exists, contains PID that belongs to a running Nginx master process.

### Related Issues

#3160 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


abom:
What's the behavior if we removed the check cmd? I think it would do this check and see if the process is running, can we test this?

sameh-farouk:
> What's the behavior if we removed the check cmd? I think it would do this check and see if the process is running, can we test this?
@abom but this will return us to this?
https://github.com/threefoldtech/js-sdk/issues/3116

waleedhammam (approved):

abom (approved):

PR #3202, opened 2021-06-21
getting rid of the development dependencies from the production server

### LINKED ISSUE #3127: development dependencies should be removed in production setup
### Description

as we almost fixed the poetry installation issue #3081 with those two PR #3124 and  #3114, now let's get rid of the development dependencies from the production server. and make use of `poetry install` option `--no-dev`.

- this will make the server slim as possible.
- will shorten the time taken to install or update the server.

### Implementation

we will need to update entrypoint.sh, Dockerfile and update.sh

sameh-farouk:
### Description

getting rid of the development dependencies from the production server.
- making the production server as slimmer as possible by removing unneeded development dependencies.
- shortening the time taken to install or update the server.
- securing the production env a bit and reducing the attack surface by installing only the necessary packages.

### Changes

- updating `entrypoint.sh`, `Dockerfile` and `update.sh` to use --no-dev option in poetry install command.

- other changes in Dockerfiles:
     - the default value for `TRC` build arg now is true.
     - copying all files in vdc-services/ to  /etc/zinit/ instead of picking up velero.yaml.
     - removing code-server installation script from the VDC image.

### Related Issues

closes #3127

### Checklist

- [X] [Pre-commit hook is installed](https://github.com/threefoldtech/js-ng#pre-commit) to do formatting checks before committing code...etc
- [ ] Tests included
- [X] Build pass
- [ ] Documentation
- [X] Code format and docstrings

### DIFF
--- a/.github/workflows/build_image.yml
+++ b/.github/workflows/build_image.yml
@@ -3,6 +3,8 @@ name: Build docker image
 on:
   push:
     branches: [ development ]
+    paths: [ jumpscale/install/Dockerfile ]
+
 jobs:
     build-docker:
         runs-on: ubuntu-latest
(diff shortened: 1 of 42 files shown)


sameh-farouk:
should be in the testing phase, but VDC deployments are broken atm due to another issue
https://github.com/threefoldtech/js-sdk/issues/3225

AhmedHanafy725 (approved):

PR #3205, opened 2021-06-21
Use different chatUID for chatflows with query params

AhmedHanafy725:
### Description

If there is a chatflow (example: extend pools) that takes query params `pool_id` to be extended and this chat is opened with a different `pool_id`, it will open the same chatflow as both depend only on the chat name.
This should solve the problem by taking the query params into consideration and get different sessions for the difference in query params.

### DIFF
--- a/jumpscale/packages/chatflows/frontend/App.vue
+++ b/jumpscale/packages/chatflows/frontend/App.vue
@@ -104,7 +104,15 @@
         return window.self === window.top
       },
       chatUID () {
-        return `${this.package}_${this.chat}`
+        let uid = `${this.package}_${this.chat}`
+        const query = this.$route.query || {}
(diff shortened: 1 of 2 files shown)


jumpscale/packages/chatflows/frontend/App.vue
```diff
@@ -104,7 +104,15 @@
         return window.self === window.top
       },
       chatUID () {
-        return `${this.package}_${this.chat}`
+        let uid = `${this.package}_${this.chat}`
+        const query = this.$route.query || {}
+        let keys = Object.keys(query)
```
xmonader, line 109: why not 
uid += keys.join("_") ?

xmonader (approved):

Issue #3171, opened 2021-06-09
Certbot cronjob not installed after moving certbot from apt to pip


MElborolossy:
Certbot cronjob not installed after moving certbot from apt to pip
Cronjob file path: `/etc/cron.d/certbot`
Cronjob file content:
```
# /etc/cron.d/certbot: crontab entries for the certbot package
#
# Upstream recommends attempting renewal twice a day
#
# Eventually, this will be an opportunity to validate certificates
# haven't been revoked, etc.  Renewal will only occur if expiration
# is within 30 days.
#
# Important Note!  This cronjob will NOT be executed if you are
# running systemd as your init system.  If you are running systemd,
# the cronjob.timer function takes precedence over this cronjob.  For
# more details, see the systemd.timer manpage, or use systemctl show
# certbot.timer.
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

0 */12 * * * root test -x /usr/bin/certbot -a \! -d /run/systemd/system && perl -e 'sleep int(rand(43200))' && certbot -q renew
```

abom:
This cron job only runs `certbot renew`, and because threebot can run certbot to:

* Fetch and install the certificate by doing `certbot run`
* Or only install the certificate [if cert paths are set](https://github.com/threefoldtech/js-sdk/blob/8354b5afb118a29943ebe1fa0122d3dcf7ac844e/jumpscale/sals/nginx/nginx.py#L319) by doing `certbot install`

In the second case, certbot won't manage this certificate and doing `certbot certificates` won't list it, hence `certbot renew` won't renew it.

IMO, we add a job for a script which ensures certbot manages this certificate, otherwise, it should run `certbot run` to fetch a new one that can be renewed later.

PR #3214, opened 2021-06-24
Implement certbot cronjob

MElborolossy:
### Description

Implement Certbot cronjob

### Changes

- Create certbot cronjob custom script, to check if there is any certificate managed by certbot or not
- Issue a new certificate if there no managed certificates
- Otherwise, run the renew command

### Related Issues

#3171

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


jumpscale/sals/nginx/nginx.py
```diff
@@ -318,16 +326,45 @@ def generate_certificates(self, retries=6):
...
+
+    def obtain_and_install_certifcate(self, retries=6):
+        """Construct and Execute run certificate command,This will issue a new certificate managed by Certbot
+        Alternative to certbot run
+
+        Args:
+            retries (int, optional): Number of retries Certbot will try to install the certificate if failed. Defaults to 6.
+        """
+        # To make sure that the old info clear
+        self.certbot.key_path = ""
```
abom, line 357: I think no need to alter current object state by removing key and cert paths, Certbot run command will ignore these path arguments anyway.

```suggestion
```

jumpscale/sals/nginx/nginx.py
```diff
@@ -318,16 +326,45 @@ def generate_certificates(self, retries=6):
...
+    def install_certifcate(self, retries=6):
+        """Construct and Execute install certificate command
+        Alternative to certbot install
+
+        Args:
+            retries (int, optional): Number of retries Certbot will try to install the certificate if failed. Defaults to 6.
+        """
+        cmd = self.certbot.install_cmd
+        j.logger.debug(f"Execute: {' '.join(cmd)}")
+        for _ in range(retries):
```
abom, line 342: do we have a generic way to do retries? if not we can add a function or a decorator that can do this and reuse it in both `install_certificate` and `obtain_and_install_certificate`.

jumpscale/install/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN if [ "$TRC" = "true" ] ; then wget https://github.com/threefoldtech/tcproute
     && chmod +x /sbin/trc ; fi
 
 COPY rootfs /
+COPY certbot/certbot /etc/cron.d
```
abom, line 25: can be renamed to certbot/cronjob?

jumpscale/sals/nginx/nginx.py
abom: also, I think we can remove this retries here, we did it for run command because the CA server might timeout...etc, in case of installation, this shouldn't happen.

MElborolossy: Do we still need the function or decorator after using it in `obtain_and_install_certificate` only?

jumpscale/sals/nginx/nginx.py
MElborolossy: Ok, I will remove them in the next commit

jumpscale/install/Dockerfile
MElborolossy: Your idea is better to make it clear in the source folder, but this name will not be clear in the destination folder, right?

jumpscale/sals/nginx/nginx.py
abom: I think no, we don't need it.

jumpscale/install/Dockerfile
abom: we can name the destination file as needed

MElborolossy: Ok, I will modify it in the next commit.

abom (approved):

PR #3235, opened 2021-07-06
Runs the js-sdk image build workflow only if Dockerfile changed

sameh-farouk:
### Description

Runs the js-sdk image build workflows only if Dockerfile changed

### Changes

these workflows will run only if Dockerfile change instead of running on every push to development:
- .github/workflows/build_image_with_trc.yml
- .github/workflows/build_image.yml
- .github/workflows/build_vdc_image.yml

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/.github/workflows/build_image.yml
+++ b/.github/workflows/build_image.yml
@@ -3,6 +3,7 @@ name: Build docker image
 on:
   push:
     branches: [ development ]
+    paths: [ jumpscale/install/Dockerfile ]
 jobs:
     build-docker:
         runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


abom (approved):

abom:
What if it depends on other files that can be changed?

xmonader (changes requested):
this will keep all of the images outdated..

sameh-farouk:
as  discussed 
> this will keep all of the images outdated.

Issue #3296, opened 2021-09-03
creation/deletion actions no the VDC panel should reload the current VDC information


xmonader:
- Go to you VDC
- Create a VM
- Go to list VMs (you won't find the newly created VM) 

Current solution:
- need to manually refresh

Expected workflow: invalidate the current VDC loaded info and update it

RafyAmgadBenjamin:
### Done
### Branch
`development_vdc_vms_fix`
### Commit
6523e8ef45b31dd516fe5d4c10430b2edaf2fce8

![image](https://user-images.githubusercontent.com/46320202/132984383-c467cf09-4d58-43ec-87ce-14385a6a787a.png)
![image](https://user-images.githubusercontent.com/46320202/132984385-b050d342-efb6-4b60-864f-98b3d6263abd.png)
![image](https://user-images.githubusercontent.com/46320202/132984387-197f8df1-c257-45c1-9fc4-38176d3473c4.png)

This is done without refreshing the page

MohamedElmdary:
After discussing this issue with @abom we figured out that issue related to `backend`.
Somehow backend returns non consistent data (might be due to `cache`)
e.g
If you send a deletion request to `remove` a *virtual machine* => server returns back that it was deleted; however if you try to fetch all available *vms* from the server again it will return all virtual machines including the deleted one for at least `10` seconds.
Same issue applied on creation. 

Methods that we discussed about
- `get_vdc` => jumpscale/packages/vdc_dashboard/bottle/vdc_helpers.py
- `delete_vmachine` => jumpscale/packages/vdc_dashboard/bottle/api/deployments.py

xmonader:
let's have it as a known issue

PR #3313, opened 2021-09-12
VDC dashboard: fix reloading vdc info after create/delete

RafyAmgadBenjamin:
### Description

- Update virtual machine component to load its data (VMs) using created event
- while deletion we raise an event in which we remove the delete VM

### Changes

load data in created event
jumpscale/packages/vdc_dashboard/frontend/components/vdcSettings/VirtualMachine.vue
### Related Issues
https://github.com/threefoldtech/js-sdk/issues/3296

### DIFF
--- a/jumpscale/packages/vdc_dashboard/frontend/components/vdcSettings/VirtualMachine.vue
+++ b/jumpscale/packages/vdc_dashboard/frontend/components/vdcSettings/VirtualMachine.vue
@@ -88,16 +88,6 @@
 
 <script>
 module.exports = {
-  props: {
-    // vmachines: {
-    //   type: Array,
-    //   default: () => [],
-    // },
-    // tableloading: {
-    //   type: Boolean,
-    //   default: false,
-    // },
-  },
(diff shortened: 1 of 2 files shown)


jumpscale/packages/vdc_dashboard/frontend/components/vdcSettings/VirtualMachine.vue
```diff
@@ -81,30 +81,22 @@
       title="Delete VM"
       :messages="deletionMessages"
       :wid="selectedvm"
-      @reload-vdcinfo="getVMSinfo"
+      @reload-vdcinfo="UpdateVms"
```
dmahmouali, line 84: I guess the real issue is that the implementation of  `UpdateVms` does not exist. so if added in the methods and bind the handler in the home call it will work fine.

```
UpdateVms() {
      this.$emit("reload-vdcinfo", {
        message: "VDC info has been reloaded successfully!",
      });
    },
```

xmonader (approved):

PR #3345, opened 2021-11-14
Fix: using ref instead of old branch var name

sameh-farouk:


jumpscale/packages/vdc_dashboard/bottle/api/deployments.py
```diff
@@ -420,13 +420,13 @@ def accept():
 @app.route("/api/update", method="GET")
 @package_authorized("vdc_dashboard")
 def update():
-    branch_param = request.params.get("branch")
-    if branch_param:
-        ref = branch_param
+    ref_param = request.params.get("ref")
```
dmahmouali, line 423: I believe in keeping the branch option and adding the ref beside, it is helpful in the dev process. what do you think?

dmahmouali: I think the branch also will work in this case, I will approve :+1:

dmahmouali (approved):

jumpscale/packages/vdc_dashboard/bottle/api/deployments.py
sameh-farouk: ref could be a branch name or tag name, these two use cases work equally
/update?ref=development  // branch name
/update?ref=v11.5-b9  // tag name

